PR #33, opened 2015-02-15
Fix private dependencies.

### LINKED ISSUE #31: Move CLI tool dependencies to Cartfile.private
A consequence of everything being in the same repo and using Carthage is that I unnecessarily need to manage the dependencies of the CLI tool (Commandant, LlamaKit etc) despite only relying on the framework.

To clarify, in my own project that has a CLI component and depends on Commandant and SourceKittenFramework I need to use a version of Commandant that satisfies the SourceKitten CLI tool. The same thing goes for LlamaKit and others.

My framework and CLI tool are in separate repos but consumers of the framework are subject to the same thing.

I understand this is probably just following the structure of the Carthage project itself, and that there is a convenience factor involved.

/cc @jspahrsummers

paulyoung:
This pull request reflects the changes discussed in Carthage/Carthage#331 and Carthage/Carthage#332.

Fixes #31.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,3 +1,1 @@
-github "Carthage/Commandant" "master"
 github "jpsim/SwiftXPC" "master"
-github "jspahrsummers/xcconfigs" >= 0.6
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks Paul! :sparkles:

PR #34, opened 2015-02-26
Mention that git clone should be used for installation

Ben-G:
Just ran into this small issue: downloaded the master branch as zip file. Didn't see that this project depends on git submodules, so installation failed.

We could also mention that ZIP download + using carthage to install dependencies works as well, but I think that would be overkill.


jpsim:
Thanks!

PR #39, opened 2015-04-06
Added definition line ranges to declarations.

jpsim:
Needed for https://github.com/realm/jazzy/issues/198. /cc @segiddins


segiddins:
This seems fine, but maybe some unit test coverage?

PR #41, opened 2015-04-06
Parse `doc.full_as_xml`

### LINKED ISSUE #37: Parse XML
`key.doc_as_full_xml` returns XML and it'd be great if SourceKitten could parse that instead of jazzy. We'll eventually need to do this for Clang output too.

jpsim:
Fixes #37.

I'm compiling SWXMLHash into SourceKittenFramework because A) its master branch (required for Swift 1.1) doesn't have any shared schemes and B) XML parsing is only used in one place.


Source/SourceKittenFramework/SwiftDocKey.swift
```diff
@@ -11,6 +11,8 @@ import SwiftXPC
 
 /// SourceKit response dictionary keys.
 internal enum SwiftDocKey: String {
```
segiddins: keep these alphabetized within each section?

segiddins:
This is fine, but I'd like to see the `SWXMLHash` eventually moved to be external.

jpsim:
> This is fine, but I'd like to see the SWXMLHash eventually moved to be external.

It's done on its `xcode-6.3` branch and that's what we'll move to once it's in `master`: https://github.com/drmohundro/SWXMLHash/pull/26

PR #49, opened 2015-05-20
Remove LlamaKit and Commandant from SourceKittenFramework.

swizzlr:
For those of us using _just_ the framework! :heart:


jpsim:
Thanks for the PR, and I agree with the end goal of keeping the framework lean, but this approach has some issues. It's what the project was doing up until earlier this week, actually.

The problem is that because Swift frameworks must be dynamically linked, we need to keep them somewhere outside the command line binary. That place is in `/Library/Frameworks` in some cases.

So say we don't vendor Commandant and LlamaKit inside SourceKittenFramework, `/Library/Frameworks` would contain all three of those frameworks. So if say `CarthageKit` also attempts to dynamically load Commandant, it'll find `/Library/Frameworks/Commandant.framework` rather than `/Library/Frameworks/CarthageKit.framework/Frameworks/Commandant.framework` and those may be incompatible versions.

I think the proper solution would be to store the frameworks maybe in `/Library/Application Support/SourceKitten` or similar instead of in `/Library/Frameworks`.

swizzlr:
Ah, so if that last solution were added to this PR it may be acceptable?

jpsim:
I think that would be the appropriate solution, yes.

jpsim:
This shouldn't be a problem anymore.

PR #54, opened 2015-08-02
minimize Makefile hackery

jpsim:


jpsim:
FYI @mdiep you may be interested in these changes for Carthage.

PR #57, opened 2015-08-03
Better ObjC support

jpsim:
Still to do:

- [ ] <del>automatically determine compiler arguments. However, unlike Swift, we don't have to build the module prior to parsing the docs, so this should be much faster, assuming we find an alternative way to get the compiler arguments.</del>
- [x] remove unused code in `Clang+SourceKitten.swift`, `JSONOutput.swift`, `SourceDeclaration.swift`.
- [x] move SequenceType extensions into dedicated `SequenceType+SourceKitten.swift` file.
- [x] parse `#pragma mark`s instead of the `@name` comment hack.
- [x] prevent enums from being declared as both type definitions and enums
- [x] stop prepending `MARK: ` from extracted pragma marks
- [x] render `parsed_declaration_start` and `parsed_declaration_end`
- [x] make `DocLine` and `DocColumn` Ints in JSONOutput


tgoyne:
I haven't looked if you ended up fixing the same bugs, but the initial pull in from the hackathon stuff doesn't appear to have included https://github.com/realm/hackathon/compare/00fae1bafd71e5c0d013065d271f2a38b0861834...sourcekitten-objc, which fixed a bunch of issues with comment formatting.

jpsim:
Hmm, yeah it seems the version I had locally was way off. I did this on the plane without Internet access, so all I had was this:

```
* e6da944 - add newlines <Thomas Goyne>
* c4944fd - ridiculous MARK hacks <Thomas Goyne>
* 87a2428 - stuff <Thomas Goyne>
* 0071851 - better text <Thomas Goyne>
* a944f1d - better text <Thomas Goyne>
* aa5928a - handle verbatim lines <Thomas Goyne>
* 26a81b0 - better whitespace in declarations <Thomas Goyne>
* 1a275de - output multiple files correctly <Thomas Goyne>
* 28f6607 - generate docs for realm <Thomas Goyne>
* 4ae070b - format declarations slightly less badly <Thomas Goyne>
* 2a8a425 - fix <Thomas Goyne>
* 6a046d3 - stuff <Thomas Goyne>
* 7a9974b - stuff <Thomas Goyne>
* cb0e0bb - no auto bootstrap <Thomas Goyne>
* 3fc1f1d - Skip non-declaration nodes <Thomas Goyne>
* e4778d4 - Move libclang interactions to type extensions <Thomas Goyne>
* ca5a6cd - disable unrelated tests <Thomas Goyne>
```

But the comment parsing is now being done almost entirely in jazzy itself, so we may be fine with this as-is. I'll check it out.

jpsim:
Looking at the latest `sourcekitten-objc` branch, it looks like really nice work, but I believe we're better off parsing the raw comment with Jazzy's markdown parser (otherwise it'd be really tricky to support markdown).

I've archived the work in the [new-hackathon-objc](https://github.com/jpsim/SourceKitten/tree/new-hackathon-objc) branch in case we want to re-introduce that in the future.

jpsim:
> automatically determine compiler arguments.

We might be able to use https://github.com/samdmarshall/xcparse. Or even a paired down version of it that just does this.

jpsim:
Random thought: appledoc supports discussions attached to sections, but moving to `#pragma mark`s prevents us from doing that.

jpsim:
I'm giving up on rendering `@return` and `@param` in jazzy for now, since I'm finding it quite challenging to use redcarpet to extract those list items in a structured way. I'm sure there's an easy way to do it, I just haven't figured it out yet, and I don't feel like further delaying Jazzy's Objective-C support purely for a nicer implementation.

So I'm moving on to the other WIP aspects of this PR.

jpsim:
The major remaining step is automatically determining compiler arguments for Objective-C projects, which I'm inclined to do in a separate PR, merging this as-is first.

jpsim:
I just created #88 and #89 to address the two points I *wanted* to do in here, but that was getting out of scope.

PR #60, opened 2015-08-15
Add File lines convenience method

keith:
This adds a lines accessory to File. This can be useful if you want to iterate over all lines in the file multiple times.

This will be super useful in [SwiftLint](https://github.com/realm/SwiftLint) where the same file is being used over and over and the contents of that file are broken into these lines many times. If this PR looks good I'll submit one to SwiftLint with the necessary changes.

The one downside to this is the file needs to be a `var` since this property is lazy. I think that is preferred over setting the lines for every file since they won't always be used.

### DIFF
--- a/Source/SourceKittenFramework/File.swift
+++ b/Source/SourceKittenFramework/File.swift
@@ -16,6 +16,10 @@ public struct File {
     public let path: String?
     /// File contents.
     public let contents: String
+    /// File lines.
+    lazy public var lines: [Line] = {
+        return self.contents.lines()
+    }()
 
     /**
     Failable initializer by path. Fails if file contents could not be read as a UTF8 string.
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -8,6 +8,8 @@
 
 import Foundation
 
+public typealias Line = (index: Int, content: String)
+
 private let whitespaceAndNewlineCharacterSet = NSCharacterSet.whitespaceAndNewlineCharacterSet()
 
 extension NSString {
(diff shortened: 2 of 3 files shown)


keith:
Just noticed how syntaxMap is cached. If we wanted I could change this to be cached the same way.

Source/SourceKittenFramework/String+SourceKitten.swift
```diff
@@ -151,6 +151,18 @@ extension NSString {
...
     /**
+    Returns an array of Lines for each line in the file
+    */
+    internal func lines() -> [Line] {
+        var lines = [Line]()
+        var lineIndex = 1
+        enumerateLines { line, _ in
+            lines.append((lineIndex++, line))
+        }
+        return []
```
jpsim: `return lines`

keith: Ha yep!

Source/SourceKittenFramework/File.swift
```diff
@@ -10,12 +10,18 @@ import Foundation
 import SwiftXPC
 import SWXMLHash
 
+public typealias Line = (index: Int, content: String)
```
jpsim: this should probably go in `String+SourceKitten.swift` since it's generally applicable outside of `File`s.

keith: That's fair. I put it here just because this was the only place it was used. I'll move it

jpsim:
Looks good, I'm ok with having the `lines` property be lazy var. Add an entry in the changelog and it's good to go! :+1:

keith:
Awesome! I've updated the changelog and fixed the other comments!

jpsim:
Thanks Keith!

keith:
Thanks! More SwiftLint PRs incoming!

PR #67, opened 2015-09-21
fix '--file' option

yamaya:
conditions that return .failure enum will be only when it fails to read the file.


jpsim:
Thanks for the fix!!! :tada:

PR #71, opened 2015-09-25
Add compilerargs option to complete command.

yamaya:
To use the iOS SDK is "complete" command, `-target` option must also be appropriately used.
So, I have added a `compilerargs` as an option to pass any of the compiler options to SourceKit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,12 @@ options for the offset in the file/text provided:
 ]
 ```
 
+To use the iOS SDK is to specified the `compilerargs` option. The value
+specified by `compilerargs` the `--` to must be prefixed:
+```
+sourcekitten complete --text "import UIKit ; UIColor." --offset 22 --compilerargs -- '-target arm64-apple-ios9.0 -sdk /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS9.0.sdk'
+```
+
 ## Doc
 
 Running `sourcekitten doc` will pass all arguments after what is parsed to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+## 0.5.2
+
+##### Breaking
+
+None.
+
+##### Enhancements
+
+* Add `compilerargs` option to complete command.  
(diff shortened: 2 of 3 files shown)


jpsim:
I'd prefer just having a `--platform` argument instead that switches the SDK (possible values should be `iOS`, `Mac`, `watchOS`, `tvOS`)

segiddins:
@jpsim it's usually better to use the SDK instead of the platform

jpsim:
> @jpsim it's usually better to use the SDK instead of the platform

Certainly not from a usability standpoint. Can you elaborate here?

segiddins:
iOS, watchOS, and tvOS both have two SDKs for their platforms. Those SDKs are free to have entirely different contents. Additionally, when  inevitably adds yet another SDK, this value can be passed through transparently, without needing to update an enum that maps between platform and SDK

README.md
```diff
@@ -73,6 +73,12 @@ options for the offset in the file/text provided:
 ]
 ```
 
+To use the iOS SDK is to specified the `compilerargs` option. The value
```
jpsim: How about just saying this:

> Use the `compilerargs` argument to pass compiler arguments to SourceKit:

And is the `--` correct here? Normally, the convention is to treat everything past the `--` as a variable-length array of values, which is not the case here since the compiler arguments are given as a single string.

jpsim:
Sure, keeping `compilerargs` is fine then. Could you also add a changelog entry @yamaya?

README.md
yamaya: Correct .

If you do not put a `--` , it would have been interpreted as an option of SourceKitten.

For example , `-target` is would be treated as an option of sourcekitten. as a result, the unknown option error.

By the way , this feature is provided by the Commandant.

jpsim: Then are the quotes wrong in the example? I don't think they are supposed to be there.

yamaya: Quotes is required.

If you run without quotes It is an command error(`Illegal instruction:4`).
It because the shell will run the argument parsing was delimiter whitespace.

jpsim:
Huge thanks for this, @yamaya! :+1:

PR #78, opened 2015-11-01
[WIP] Update dependencies whose branches no longer exist.

paulyoung:
This updates Commandant and SWXMLHash since the branches specified in the Cartfile no longer exist.

I've marked this as WIP since a test is now failing.


paulyoung:
It looks as if the failure is due to the updated version of Commandant being different from the fixture.

jpsim:
Thanks for this @paulyoung! I've updated the Commandant fixture in #80.

PR #85, opened 2015-11-05
Link Commandant and Result with sourcekitten, not SourceKittenFramework.

paulyoung:
I was surprised to find that linking `SourceKittenFramework` to a target caused a build failure with the message `Framework not found Result`.

Both Result and Commandant are not included in the public Cartfile, and appear to be dependencies of `sourcekitten` and not `SourceKittenFramework`. These changes reflect that.

I also removed an old reference to LlamaKit.

Please review! :pray:


jpsim:
This should be fine, thanks @paulyoung!

jpsim:
Turns out this certainly was *not* fine, as Result & Commandant were no longer copied into SourceKittenFramework. I've addressed this in 63f0989a0d1f7d3b9fa0947aa8fb8a5cf6b6668b.

paulyoung:
@jpsim could you help me understand how they're used in`SourceKittenFramework`? Aren't they dependencies for `sourcekitten`?

jpsim:
Right. Essentially because using dynamic frameworks for command line tools is a mess. In order to not collide with other command line tools using Commandant and Result, I package those frameworks into SourceKittenFramework. Because these dynamic frameworks have to be installed in a shared location. :cry:

PR #92, opened 2015-11-10
fix empty Objective-C enum names

jpsim:
libClang considers enums declared like `typedef enum {} name;` rather than `NS_ENUM()`
to have a cursor spelling of "" (empty string). So we parse the USR to extract the actual name.

This fixes https://github.com/realm/jazzy/issues/342 /cc @jverkoey

### DIFF
--- a/Source/SourceKittenFramework/Clang+SourceKitten.swift
+++ b/Source/SourceKittenFramework/Clang+SourceKitten.swift
@@ -74,19 +74,6 @@ extension CXCursor {
         return ObjCDeclarationKind.fromClang(kind)
     }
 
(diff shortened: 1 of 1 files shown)


Source/SourceKittenFramework/Clang+SourceKitten.swift
```diff
@@ -74,19 +74,6 @@ extension CXCursor {
         return ObjCDeclarationKind.fromClang(kind)
     }
 
-    func mark() -> String? {
```
jpsim: I removed this because we no longer use it

PR #95, opened 2015-11-11
generate Swift interface to Objective-C header

jpsim:
super early proof of concept generating Swift interface for Objective-C header


jverkoey:
In the absence of any reasonable ability to help with this in the short term, I give you an auditorium filled with cheering pikachus.

ab749e6b99a50eb2b4b9d59cbcebd001

orta:
Here's a winking galaxy of hope too

The Universe's RW it helps me get lucky with the ladies

jpsim:
I just found out that OSS Swift includes `lib/PrintAsObjC`: https://github.com/apple/swift/blob/master/lib/PrintAsObjC/PrintAsObjC.cpp

If anyone's interested in perusing that.

jpsim:
I've rebased this and improved the test. Should be usable as-is.

jpsim:
The SPM tests are failing to build because it's trying to build the fixture, even though it's just a Swift interface, not valid Swift source. Excluding files from tests isn't supported in the Swift toolchain we're using for now, so I'll delay merging this until that's addressed.

jpsim:
PR in jazzy to integrate this: realm/jazzy#530

PR #103, opened 2015-11-20
Use `cp -Rf` instead of `cp -rf` in Makefile

norio-nomura:
`-r` option of `cp` copies symbolic links as normal files.
As the result of that, `SourceKittenFramework.framework` that distributed by Homebrew contains a lot of duplicated files inside that.
`-R` option copies symbolic links as symbolic links.

On my testing, this fix reduces the size of `sourcekitten-0.6.0.el_capitan.bottle.tar.gz` from 5.5MB to 2.1MB.


norio-nomura:
https://github.com/Carthage/Carthage/pull/932

jpsim:
Thanks again.

PR #108, opened 2015-11-27
make `File` `final class` instead of `struct`

jpsim:
`File` is now a `final class` instead of a `struct` and `contents` & `lines` are now marked as `var`. This was done so that mutations to the underlying file on disk can be mirrored in a `File` instance.

### DIFF
--- a/Source/SourceKittenFramework/File.swift
+++ b/Source/SourceKittenFramework/File.swift
@@ -11,13 +11,13 @@ import SwiftXPC
 import SWXMLHash
 
 /// Represents a source file.
-public struct File {
+public final class File {
     /// File path. Nil if initialized directly with `File(contents:)`.
     public let path: String?
     /// File contents.
-    public let contents: String
+    public var contents: String
     /// File lines.
-    public let lines: [Line]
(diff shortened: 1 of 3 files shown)


Source/SourceKittenFramework/File.swift
```diff
@@ -31,6 +31,9 @@ public struct File {
             lines = contents.lines()
         } catch {
             fputs("Could not read contents of `\(path)`\n", stderr)
+            // necessary to set contents & lines because of rdar://21744509
+            contents = ""
+            lines = []
```
keith: Yay for this being fixed in swift-next https://github.com/apple/swift/commit/f524802e129f17de16fa6f8c3f08e5bcf75cc581

jpsim: Yay for *knowing* this is fixed in swift-next! :tada:

jpsim: Could not have happened a week ago.

paulyoung: :heart:

PR #111, opened 2015-12-04
Fix indexOfByteOffset()

norio-nomura:


jpsim:
This is great, thanks! PS: I'm still reviewing your other PRs for the rpaths, they look good, but I prefer being extra cautious with those because touching them in the past has caused issues.

norio-nomura:
:smile:
There's no hurry about that.

Issue #87, opened 2015-11-06
Making SwiftDocKey and its methods public


paulyoung:
I think this would be useful. Are there any reasons not to do this, or an alternative that doesn't require it?

jpsim:
The reason why these aren't public is that none of the current consumers of SourceKitten (jazzy & SwiftLint) need access to them.

If I knew what your interest and goals with SourceKitten were, I could better decide if the best course is to just make these public, or to expose the information from them some other way.

paulyoung:
For example:

```
let docs = SwiftDocs(file: file, arguments: arguments)
let dict = toAnyObject(docs.docsDictionary)
```

In this case I'd then need to resort to `dict["key.substructure"]` instead of `dict[SwiftDocKey.Substructure.rawValue]` or even better, `SwiftDocKey.getSubstructure(dict)`.

paulyoung:
To provide some more context, I'm currently interested in the following:

- `key.parsed_declaration`
- `key.name`
- `key.typename`
- `key.accessibility`

I know some of these _don't_ have methods on `SwiftDocKey` so would be happy to contribute those _or_ just create a local `extension`, provided that `get` was also made `public`.

jpsim:
I wonder if serializing the `XPCDictionary`s `SwiftDoc` structs would be preferable to continuing with this largely untyped approach?

paulyoung:
I think you're proposing something more like:

```swift
public enum Accessibility: String {
    case Internal = "source.lang.swift.accessibility.internal"
    case Private = "source.lang.swift.accessibility.private"
    case Public = "source.lang.swift.accessibility.public"
}

public struct Substructure {
    let accessibility: Accessibility
    let name: String
    let parsedDeclaration: String
    let typeName: String
    ...
}
```

and so on.

jpsim:
Yup.

paulyoung:
This is the direction I was heading in. Would much prefer that.

paulyoung:
I'm planning on doing something around this, just been busy :bee:

paulyoung:
I worked on this a bit last night by simply taking the JSON fixtures and creating new structs based on their content.

This morning I realize that most of this is already represented but not all of it in the same 1-1 mapping. In short, I think the same thing could be achieved adding some properties to the existing `Documentation` and `SourceDeclaration` types instead.

@jpsim does that sound OK to you?

paulyoung:
[...]

PR #112, opened 2015-12-04
[WIP] Typed declarations.

paulyoung:
This relates to #87.

I've added `FIXME`s where I ran into hurdles that didn't have an easy immediate fix (or I wasn't sure what to do) just to get something started.

### DIFF
--- a/Source/SourceKittenFramework/DeclarationType.swift
+++ b/Source/SourceKittenFramework/DeclarationType.swift
@@ -10,7 +10,7 @@ public protocol DeclarationType {
     var language: Language { get }
     var kind: DeclarationKindType? { get }
     var location: SourceLocation { get }
-    var extent: Range<SourceLocation> { get } // FIXME: Type 'SourceLocation' does not conform to protocol 'ForwardIndexType'
+    var extent: (start: SourceLocation, end: SourceLocation) { get }
     var name: String? { get }
     var typeName: String? { get }
     var usr: String? { get }
--- a/Source/SourceKittenFramework/DeclarationType.swift
+++ b/Source/SourceKittenFramework/DeclarationType.swift
@@ -1,0 +1,20 @@
+//
+//  DeclarationType.swift
+//  SourceKitten
+//
+//  Created by Paul Young on 12/4/15.
+//  Copyright © 2015 SourceKitten. All rights reserved.
+//
+
+public protocol DeclarationType {
+    var language: Language { get }
+    var kind: DeclarationKindType? { get }
+    var location: SourceLocation { get }
+    var extent: Range<SourceLocation> { get } // FIXME: Type 'SourceLocation' does not conform to protocol 'ForwardIndexType'
+    var name: String? { get }
+    var typeName: String? { get }
+    var usr: String? { get }
+    var declaration: String? { get }
+    var documentationComment: String? { get }
+    var children: [DeclarationType] { get }
+}
--- a/Source/SourceKittenFramework/ObjCDeclaration.swift
+++ b/Source/SourceKittenFramework/ObjCDeclaration.swift
@@ -12,7 +12,7 @@ public struct ObjCDeclaration: DeclarationType {
     public let language: Language = .Swift
     public let kind: ObjCDeclarationKind? // FIXME: Type 'ObjCDeclaration' does not conform to protocol 'DeclarationType'
     public let location: SourceLocation
-    public let extent: Range<SourceLocation> // FIXME: Type 'SourceLocation' does not conform to protocol 'ForwardIndexType'
+    public let extent: (start: SourceLocation, end: SourceLocation)
     public let name: String?
     public let typeName: String?
     public let usr: String?
--- a/Source/SourceKittenFramework/ObjCDeclaration.swift
+++ b/Source/SourceKittenFramework/ObjCDeclaration.swift
@@ -1,0 +1,110 @@
+//
+//  ObjCDeclaration.swift
+//  SourceKitten
+//
+//  Created by Paul Young on 12/4/15.
+//  Copyright © 2015 SourceKitten. All rights reserved.
+//
+
+import Foundation
+
+public struct ObjCDeclaration: DeclarationType {
+    public let language: Language = .Swift
+    public let kind: ObjCDeclarationKind? // FIXME: Type 'ObjCDeclaration' does not conform to protocol 'DeclarationType'
+    public let location: SourceLocation
+    public let extent: Range<SourceLocation> // FIXME: Type 'SourceLocation' does not conform to protocol 'ForwardIndexType'
+    public let name: String?
+    public let typeName: String?
+    public let usr: String?
+    public let declaration: String?
+    public let documentationComment: String?
+    public let children: [DeclarationType]
+    
+    /// Returns the USR for the auto-generated getter for this property.
+    ///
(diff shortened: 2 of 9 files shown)


Source/SourceKittenFramework/ObjCDeclaration.swift
```diff
@@ -0,0 +1,110 @@
...
+//
+//  Created by Paul Young on 12/4/15.
+//  Copyright © 2015 SourceKitten. All rights reserved.
+//
+
+import Foundation
+
+public struct ObjCDeclaration: DeclarationType {
+    public let language: Language = .Swift
+    public let kind: ObjCDeclarationKind? // FIXME: Type 'ObjCDeclaration' does not conform to protocol 'DeclarationType'
```
paulyoung: This is because the expected type here `DeclarationKindType`.

Source/SourceKittenFramework/ObjCDeclaration.swift
```diff
@@ -0,0 +1,110 @@
...
+        lhs.location == rhs.location
+}
+
+// MARK: Comparable
+
+/// A [strict total order](http://en.wikipedia.org/wiki/Total_order#Strict_total_order)
+/// over instances of `Self`.
+public func <(lhs: ObjCDeclaration, rhs: ObjCDeclaration) -> Bool {
+    return lhs.location < rhs.location
+}
```
paulyoung: My intention was for `Hashable` / `Comparable` conformance to be on `DeclarationType`. Let me take a stab at that.

paulyoung: I think because of the use of `Self` in `Equatable`, we'd need `AnyDeclarationType`.

Source/SourceKittenFramework/DeclarationType.swift
```diff
@@ -0,0 +1,20 @@
...
+//
+//  Created by Paul Young on 12/4/15.
+//  Copyright © 2015 SourceKitten. All rights reserved.
+//
+
+public protocol DeclarationType {
+    var language: Language { get }
+    var kind: DeclarationKindType? { get }
+    var location: SourceLocation { get }
+    var extent: Range<SourceLocation> { get } // FIXME: Type 'SourceLocation' does not conform to protocol 'ForwardIndexType'
```
jpsim: If `SourceLocation` conformed to `ForwardIndexType`, creating a `Range` with it would be very expensive. So I think `extent` should be a `(start: SourceLocation, end: SourceLocation)` tuple.

Source/SourceKittenFramework/ObjCDeclaration.swift
paulyoung: I did this in https://github.com/paulyoung/sourcekitten/pull/1.

Source/SourceKittenFramework/SwiftDeclaration.swift
```diff
@@ -0,0 +1,59 @@
...
+    public let usr: String?
+    public let declaration: String?
+    public let documentationComment: String?
+    public let children: [DeclarationType]
+    public let accessibility: Accessibility?
+}
+
+extension SwiftDeclaration {
+    public init(dictionary: XPCDictionary) {
+        kind = SwiftDocKey.getKind(dictionary).flatMap { SwiftDeclarationKind(rawValue: $0) } // FIXME: why doesn't .flatMap(SwiftDeclarationKind.init) work here?
```
paulyoung: I'm sure I've done this elsewhere. Maybe it doesn't work for `enum`s.

Source/SourceKittenFramework/SwiftDocKey.swift
```diff
@@ -62,17 +62,17 @@ internal enum SwiftDocKey: String {
...
     case USR                  = "key.usr"
-    /// Result discussion documentation of documented token (XPCArray).
+    /// Parsed declaration (String).
     case ParsedDeclaration    = "key.parsed_declaration"
-    /// Type of documented token (String).
+    /// Parsed scope end (Int64).
     case ParsedScopeEnd       = "key.parsed_scope.end"
-    /// USR of documented token (String).
+    /// Parsed scope start (Int64).
     case ParsedScopeStart     = "key.parsed_scope.start"
```
paulyoung: I can move this to another PR if necessary.

Source/SourceKittenFramework/DeclarationType.swift
paulyoung: Cool. I think it's a tuple in the existing `SourceDeclaration` but I was following your lead in https://github.com/jpsim/SourceKitten/issues/87#issuecomment-159002802.

jpsim: yeah, sorry for the mixed signals. I didn't fully think through the implications of using a `Range` when I proposed that.

Source/SourceKittenFramework/SwiftDeclaration.swift
jpsim: yeah, that's odd, but not a big deal.

Source/SourceKittenFramework/DeclarationType.swift
paulyoung: Presumably then we need another constructor for `SourceLocation`? Since the existing one takes a `CXSourceLocation`.

jpsim: Isn't there an initializer for passing in the values directly? Yes, we'll need another initializer if we want to use the same struct for Swift declarations.

paulyoung: I think there's no longer a default `struct` initializer because one was defined.

paulyoung: Nope, I was wrong. My mistake.

Source/SourceKittenFramework/SwiftDeclaration.swift
```diff
@@ -0,0 +1,75 @@
...
+            }
+                
+            if let parsedScopeStart = SwiftDocKey.getParsedScopeStart(dictionary).map({ UInt32($0) }),
+                parsedScopeEnd = SwiftDocKey.getParsedScopeEnd(dictionary).map({ UInt32($0) }) {
+        
+                let start = SourceLocation.init(file: file, line: line, column: column, offset: parsedScopeStart)
+                let end = SourceLocation.init(file: file, line: line, column: column, offset: parsedScopeEnd)
+                extent = (start: start, end: end)
+            }
+        }
```
paulyoung: I'm not entirely sure that this is the right thing to do here.

paulyoung: FWIW `.map(UInt32.init)` didn't work here ¯\\\_(ツ)\_/¯

paulyoung:
@jpsim any more thoughts on this since we last spoke? Would love to keep moving forward if we can.

jpsim:
> @jpsim any more thoughts on this since we last spoke? Would love to keep moving forward if we can.

Yes, sorry for the delay. What about our conversation about having a single `Declaration` type with a language switch property?

paulyoung:
I'll try something to that effect later tonight.

paulyoung:
I took a shot at this in https://github.com/paulyoung/sourcekitten/pull/2.

jpsim:
We'll probably never revive this as-is, closing. #428 shows promise though and is along these lines.

Thanks anyway Paul!

PR #115, opened 2015-12-09
Objective-C marks and non-ASCII characters

### LINKED ISSUE #114: Non-ASCII characters in Objective-C produce incorrect output
If I run `sourcekitten doc --objc MGLMapView.h -x objective-c -isysroot $(xcrun --show-sdk-path) -I $(pwd)` with the following (stripped-down) contents of MGLMapView.h, I get the output I expect:

```objc
#import <Foundation/Foundation.h>

/**
 ’’’
 */
@interface MGLMapView : NSObject

#pragma mark Initializing a Map View

/**
 Initializes and returns a newly allocated map view with the specified frame and the default style.
 */
- (instancetype)initWithFrame:(NSRect)frame;

#pragma mark Manipulating the Visible Portion of the Map

/**
 The coordinate bounds visible in the receiver’s viewport.
 */
@property (nonatomic) int visibleCoordinateBounds;

#pragma mark Converting Map Coordinates

/**
 Converts a point in the specified view’s coordinate system to a map coordinate.
 */
- (int)convertPoint:(NSPoint)point toCoordinateFromView:(nullable NSObject *)view;

/**
 Converts a map coordinate to a point in the specified view.
 
 @param view The view in whose coordinate system you want to locate the specified map coordinate. If this parameter is `nil`, the returned point is specified in the window’s coordinate system. If `view` is not `nil`, it must belong to the same window as the map view.
 */
- (CGPoint)convertCoordinate:(int)coordinate toPointToView:(nullable NSObject *)view;

#pragma mark Styling the Map

/**
 URL of the style currently displayed in the receiver.
 
 The URL may be a full HTTP or HTTPS URL or a Mapbox URL indicating the style’s map ID (`mapbox://styles/my_user_name/abcd1234`).
 */
@property (nonatomic, null_resettable) NSURL *styleURL;

/**
 Returns a Boolean value indicating whether the style class with the given identifier is currently active.
 */
- (BOOL)hasStyleClass:(NSString *)styleClass;

#pragma mark Annotating the Map

/**
 The complete list of annotations associated with the receiver. (read-only)
 */
@property (nonatomic, readonly, nullable) NSArray *annotations;

#pragma mark Annotating the Map 2

/**
 The complete list of annotations associated with the receiver. (read-only)
 */
@property (nonatomic, readonly, nullable) NSArray *annotations2;

@end
```

However, if I add one more Unicode character (such as `’` or `≠`) to the MGLMapView class’s documentation comment, `#pragma mark` sections go missing and character indices get knocked out of place:

```diff
8c8
<           "key.doc.comment" : "’’’",
---
>           "key.doc.comment" : "’’’’",
117c117
<               "key.doc.line" : 36,
---
>               "key.doc.line" : 35,
119c119
<               "key.parsed_scope.start" : 36,
---
>               "key.parsed_scope.start" : 35,
121c121
<               "key.parsed_scope.end" : 36,
---
>               "key.parsed_scope.end" : 35,
153,162d152
<               "key.kind" : "sourcekitten.source.lang.objc.mark",
<               "key.doc.file" : "\/Users\/mxn\/Desktop\/MGLMapView.h",
<               "key.doc.line" : 49,
<               "key.name" : "Annotating the Map",
<               "key.parsed_scope.start" : 49,
<               "key.doc.column" : 1,
<               "key.parsed_scope.end" : 49,
<               "key.filepath" : "\/Users\/mxn\/Desktop\/MGLMapView.h"
<             },
<             {
179c169
<               "key.doc.line" : 56,
---
>               "key.doc.line" : 55,
181c171
<               "key.parsed_scope.start" : 56,
---
>               "key.parsed_scope.start" : 55,
183c173
<               "key.parsed_scope.end" : 56,
---
>               "key.parsed_scope.end" : 55,
```

As far as I can tell, the issue seems to be that `String.pragmaMarks(_:excludeRanges:limitRange:)` in String+SourceKitten.swift works with NSString indices, whereas `clang_getSpellingLocation()` appears to be reporting byte offsets.

/cc @friedbunny

1ec5:
Fixes #114.

### DIFF
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -88,6 +88,24 @@ extension NSString {
     }
 
     /**
+    Converts an `NSRange` suitable for filtering `self` as an
+    `NSString` to a range of byte offsets in `self`.
+
+    - parameter start: Starting character index in the string.
+    - parameter length: Number of characters to include in range.
+
+    - returns: An equivalent `NSRange`.
+    */
+    public func NSRangeToByteRange(start start: Int, length: Int) -> NSRange? {
+        let string = self as String
+        return string.byteOffsetAtIndex(start).flatMap { stringStart in
+            return string.byteOffsetAtIndex(start + length).map { stringEnd in
+                return NSRange(location: stringStart, length: stringEnd - stringStart)
+            }
+        }
+    }
+
+    /**
     Returns a substring with the provided byte range.
 
     - parameter start: Starting byte offset.
@@ -181,6 +199,17 @@ extension String {
         return utf8.startIndex.advancedBy(offset).samePositionIn(utf16).map(utf16.startIndex.distanceTo)
     }
 
+    /**
+     Byte offset equivalent to UTF16 index.
+     
+     - parameter index: UTF16 index.
+     
+     - returns: Byte offset, if any.
+     */
+    private func byteOffsetAtIndex(index: Int) -> Int? {
+        return utf16.startIndex.advancedBy(index).samePositionIn(utf8).map(utf8.startIndex.distanceTo)
+    }
+
     /// Returns the `#pragma mark`s in the string.
     /// Just the content; no leading dashes or leading `#pragma mark`.
     public func pragmaMarks(filename: String, excludeRanges: [NSRange], limitRange: NSRange?) -> [SourceDeclaration] {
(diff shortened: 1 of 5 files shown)


Source/SourceKittenFramework/String+SourceKitten.swift
```diff
@@ -205,9 +234,13 @@ extension String {
             if markString.isEmpty {
                 return nil
             }
+            let markByteRange = self.NSRangeToByteRange(start: markRange.location, length: markRange.length)
+            if markByteRange == nil {
```
segiddins: This could be a `guard`

jpsim: Agreed.

1ec5:
Fixed in 4243ed807f4e0f251b413dced7e1e30fedab6095.

Source/SourceKittenFramework/String+SourceKitten.swift
1ec5: Fixed in 4243ed807f4e0f251b413dced7e1e30fedab6095. Thanks for the suggestion.

jpsim:
This is really great @1ec5! Thanks for the fix, and especially the nice test case. I'll add a changelog entry to describe the fix and attribute you, but that's something you should do yourself next time!

Issue #119, opened 2015-12-13
Performance regression on 0.7.0


norio-nomura:
Sorry, I have introduced performance regression by #111.
I think `func byteRangeToNSRange(start start: Int, length: Int) -> NSRange?` needs caching result.

norio-nomura:
Or `func indexOfByteOffset(offset: Int) -> Int?`.

norio-nomura:
I'm sorry, the performance regression does not caused by #111.
But 0.7.0 is so slow.
e.g. [KeychainAccess](https://github.com/kishikawakatsumi/KeychainAccess/)

norio-nomura:
I thought that the performance regression has been introduced by realm/SwiftLint#247 and filed PRs (realm/SwiftLint#263, #120 and #122) for resolving that.
SwiftLint 0.5.1 is still slower than 0.4.0, but now I close this issue.

PR #120, opened 2015-12-13
Optimize `NSRange` conversion

norio-nomura:
By applying this, duration of linting [KeychainAccess](https://github.com/kishikawakatsumi/KeychainAccess/) by SwiftLint is reduced from 27sec to 21sec.

Related: https://github.com/jpsim/SourceKitten/issues/119

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## Master
+
+##### Enhancements
+
+* Optimize `NSRange` operation.  
+  [Norio Nomura](https://github.com/norio-nomura)
+  [#120](https://github.com/jpsim/SourceKitten/pull/120)
+
 ## 0.7.0
 
 ##### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Optimize `NSRange` operation.  
   [Norio Nomura](https://github.com/norio-nomura)
-  [#120](https://github.com/jpsim/SourceKitten/pull/120)
+  [#119](https://github.com/jpsim/SourceKitten/issues/119)
 
 ## 0.7.0
 
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -80,11 +80,17 @@ extension NSString {
     */
     public func byteRangeToNSRange(start start: Int, length: Int) -> NSRange? {
         let string = self as String
-        return string.indexOfByteOffset(start).flatMap { stringStart in
-            return string.indexOfByteOffset(start + length).map { stringEnd in
-                return NSRange(location: stringStart, length: stringEnd - stringStart)
-            }
+        let startUTF8Index = string.utf8.startIndex.advancedBy(start)
+        let endUTF8Index = startUTF8Index.advancedBy(length)
+        
+        guard let startUTF16Index = startUTF8Index.samePositionIn(string.utf16),
+            let endUTF16Index = endUTF8Index.samePositionIn(string.utf16) else {
+                return nil
         }
+        
+        let location = string.utf16.startIndex.distanceTo(startUTF16Index)
+        let length = startUTF16Index.distanceTo(endUTF16Index)
+        return NSRange(location: location, length: length)
     }
 
     /**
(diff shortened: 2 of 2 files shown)


Source/SourceKittenFramework/String+SourceKitten.swift
```diff
@@ -80,11 +80,17 @@ extension NSString {
...
     public func byteRangeToNSRange(start start: Int, length: Int) -> NSRange? {
         let string = self as String
-        return string.indexOfByteOffset(start).flatMap { stringStart in
-            return string.indexOfByteOffset(start + length).map { stringEnd in
-                return NSRange(location: stringStart, length: stringEnd - stringStart)
-            }
+        let startUTF8Index = string.utf8.startIndex.advancedBy(start)
+        let endUTF8Index = startUTF8Index.advancedBy(length)
+        
+        guard let startUTF16Index = startUTF8Index.samePositionIn(string.utf16),
```
jpsim: are `utf16` views cached in the standard library? If not, we should cache it here (e.g. just calculate it once).

Source/SourceKittenFramework/String+SourceKitten.swift
```diff
@@ -98,11 +104,17 @@ extension NSString {
...
     public func NSRangeToByteRange(start start: Int, length: Int) -> NSRange? {
         let string = self as String
-        return string.byteOffsetAtIndex(start).flatMap { stringStart in
-            return string.byteOffsetAtIndex(start + length).map { stringEnd in
-                return NSRange(location: stringStart, length: stringEnd - stringStart)
-            }
+        let startUTF16Index = string.utf16.startIndex.advancedBy(start)
+        let endUTF16Index = startUTF16Index.advancedBy(length)
+        
+        guard let startUTF8Index = startUTF16Index.samePositionIn(string.utf8),
```
jpsim: same thing about caching `utf8` views.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## Master
+
+##### Enhancements
+
+* Optimize `NSRange` operation.  
+  [Norio Nomura](https://github.com/norio-nomura)
+  [#120](https://github.com/jpsim/SourceKitten/pull/120)
```
jpsim: Please refer to the issues this addresses (#119) rather than this PR number (as explained in https://github.com/jpsim/SourceKitten/blob/master/CONTRIBUTING.md).

jpsim:
Thanks for doing this!

Source/SourceKittenFramework/String+SourceKitten.swift
norio-nomura: At first since I filed https://github.com/jpsim/SourceKitten/issues/119, I tested caching full results of offsets. But that was not efficient.
By https://github.com/realm/SwiftLint/pull/263, times of calling `byteRangeToNSRange()` is omitted  on bottle neck case, so I did not include that.

jpsim: that's not what I mean. I mean are `utf16` views on Strings cached by the Swift standard library?

e.g. is it more efficient to do this:

```swift
let utf16View = string.utf16
guard let startUTF16Index = startUTF8Index.samePositionIn(utf16View),
      let endUTF16Index = endUTF8Index.samePositionIn(utf16View) else {
        return nil
}
```

or is that equivalent to this (current code):


```swift
guard let startUTF16Index = startUTF8Index.samePositionIn(string.utf16),
      let endUTF16Index = endUTF8Index.samePositionIn(string.utf16) else {
        return nil
}
```

Source/SourceKittenFramework/String+SourceKitten.swift
norio-nomura: Maybe caching exact results will not be efficient for most use case (I don't know other than SwiftLint) because inputs are matching results and those are almost unique.

But caching latest index might be efficient for SwiftLint.
I will test that.

jpsim: I don't mean caching the results of this method, but rather storing `utf8` in a variable to reuse it here, rather than calling `string.utf8` twice.

Source/SourceKittenFramework/String+SourceKitten.swift
norio-nomura: Oh, I understand.

norio-nomura: I tested with changes of that and got same performance.
It seems Swift optimizer is clever enough. :smiley: 
But apply them.

norio-nomura:
Thank you for reviews!
I will make another PR about the caching that I mentioned on `NSRangeToByteRange`, if that will be efficient.

jpsim:
> I will make another PR about the caching that I mentioned on NSRangeToByteRange, if that will be efficient.

I'm not sure if the gains there will be worth it, but if they're 10% or more, it'd be great.

PR #122, opened 2015-12-14
Optimize `NSString.NSRangeToByteRange(start:length:)`

norio-nomura:
Introduce `ByteOffsetCache` for calculating byte offset.
By applying this, linting KeychainAccess by SwiftLint-0.5.1 is reduced from:
```
swiftlint  21.87s user 0.24s system 97% cpu 22.611 total
```
to:
```
swiftlint  16.62s user 0.24s system 97% cpu 17.362 total
```

On Instruments, `NSRangeToByteRange()` is reduced from 6579ms:
screenshot 2015-12-14 11 38 48

 to 315ms:
screenshot 2015-12-14 11 39 43

 on that test.

### DIFF
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -21,7 +21,66 @@ private let commentLinePrefixCharacterSet: NSCharacterSet = {
   return characterSet
 }()
 
+private var keyByteOffsetCache = "ByteOffsetCache"
+
 extension NSString {
+    /**
+    ByteOffsetCache is cache byte offset and UTF8Index of UTF16 based location.
+    This depends on using for that will be calculated from begening of the string to the end.
+    */
+    @objc private class ByteOffsetCache: NSObject {
+        struct Entry {
+            let byteOffset: Int
+            let index: String.UTF8Index
+        }
+        
+        var cache = Dictionary<Int, Entry>()
+        let utf8View: String.UTF8View
+        
+        init(_ string: String) {
+            self.utf8View = string.utf8
+        }
+        
+        func byteOffsetFromLocation(location: Int, andIndex index: String.UTF8Index) -> Int {
+            if let entry = cache[location] {
+                return entry.byteOffset
+            } else {
+                let entry: Entry
+                if let nearestLocation = cache.keys.filter({ $0 < location }).maxElement() {
+                    let nearestEntry = cache[nearestLocation]!
+                    let byteOffset = nearestEntry.byteOffset + nearestEntry.index.distanceTo(index)
+                    entry = Entry(byteOffset: byteOffset, index: index)
+                } else {
+                    let byteOffset = utf8View.startIndex.distanceTo(index)
+                    entry = Entry(byteOffset: byteOffset, index: index)
+                }
+                cache[location] = entry
+                
+                // keep largest 5 entry
+                if cache.keys.count > 5 {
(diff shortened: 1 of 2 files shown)


norio-nomura:
This is that I mentioned in https://github.com/jpsim/SourceKitten/pull/120.

norio-nomura:
I think the performance issue that related to https://github.com/realm/SwiftLint/pull/247 is almost resolved, and another performance issues that made SwiftLint slower are caused by rules.

norio-nomura:
Today, I worried about that this PR would make leaks String instance.
But, it did not happen as I tested using Instruments. :relieved:

Source/SourceKittenFramework/String+SourceKitten.swift
```diff
@@ -21,7 +21,66 @@ private let commentLinePrefixCharacterSet: NSCharacterSet = {
...
+                    let byteOffset = nearestEntry.byteOffset + nearestEntry.index.distanceTo(index)
+                    entry = Entry(byteOffset: byteOffset, index: index)
+                } else {
+                    let byteOffset = utf8View.startIndex.distanceTo(index)
+                    entry = Entry(byteOffset: byteOffset, index: index)
+                }
+                cache[location] = entry
+                
+                // keep largest 5 entry
+                if cache.keys.count > 5 {
```
jpsim: is it really necessary to evict cache entries? An Int/Int keypair is minuscule in memory compared to the rest of the strings we're working with.

Source/SourceKittenFramework/String+SourceKitten.swift
```diff
@@ -21,7 +21,66 @@ private let commentLinePrefixCharacterSet: NSCharacterSet = {
...
 
+private var keyByteOffsetCache = "ByteOffsetCache"
+
 extension NSString {
+    /**
+    ByteOffsetCache is cache byte offset and UTF8Index of UTF16 based location.
+    This depends on using for that will be calculated from begening of the string to the end.
+    */
+    @objc private class ByteOffsetCache: NSObject {
+        struct Entry {
```
jpsim: `Entry` isn't a very descriptive name in this case. `ByteOffsetIndexPair` could work.

Source/SourceKittenFramework/String+SourceKitten.swift
norio-nomura: Now I tested again.
I confirmed that evicting cache does not affect performance. :+1: 
I'll remove that.

norio-nomura:
Thank you for reviews.
I updated.

norio-nomura:
> Today, I worried about that this PR would make leaks String instance.
> But, it does not happen as I tested using Instruments. :relieved:

I also found that using `NSString.byteOffsetCache` cause memory leak if `NSString` is native `NSString`.
If `NSString` is casted from `String`, that does not cause memory leak.

Would we avoid using this approach?

norio-nomura:
It seems I need to using two workarounds for resolving memory leaks.
1. Check `NSString` is `NSTaggedPointerString`. If length is short, don't use cache.
2. Create new instance of string for avoiding Circular reference.

norio-nomura:
I added workarounds for memory leaks.

norio-nomura:
http://www.openradar.me/radar?id=5050693400395776

norio-nomura:
rebased to master

jpsim:
Awesome! Thanks for yet another solid contribution.

norio-nomura:
:smile:

PR #124, opened 2015-12-18
Make `String.commentBody(range:)` emoji safe

norio-nomura:


jpsim:
Oops, this was obviously wrong. Thanks for fixing!

norio-nomura:
:+1:

PR #127, opened 2015-12-24
add NSString.lineAndCharacterForByteOffset

jpsim:


norio-nomura:
:+1: 
We need more performance optimization.

jpsim:
> We need more performance optimization.

True, but we may have to rethink the string cache altogether though. I think we may be able to combine some caching instead of essentially adding a 3rd cache set here.

norio-nomura:
I think converting all byte offset into utf16 index before passing models to client (SwiftLint).
Caching may not be needed by that.

norio-nomura:
I have a small-scale implementation idea that extends current implementation.
I'll write that.

PR #128, opened 2015-12-25
Optimize NSString.lineAndCharacterForByteOffset()

norio-nomura:
Add `NSString.CacheContainer.lineAndCharacterForByteOffset(_:)`

This commit reduces the duration of linting KeychainAccess by SwiftLint (8d529f1) from:
```
swiftlint lint  27.27s user 0.25s system 96% cpu 28.424 total
```
to:
```
swiftlint lint  3.76s user 0.10s system 89% cpu 4.291 total
```

This is that I mentioned in #127

### DIFF
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -35,13 +35,14 @@ extension NSString {
             let index: String.UTF8Index
         }
 
-        struct RangeLinePair {
+        struct RangesAndLine {
             let range: NSRange
+            let byteRange: NSRange
             let line: Line
         }
 
         var byteOffsetIndexPairs = Dictionary<Int, ByteOffsetIndexPair>()
-        let rangeLinePairs: [RangeLinePair]
+        let rangesAndLine: [RangesAndLine]
         let utf8View: String.UTF8View
 
         init(_ string: NSString) {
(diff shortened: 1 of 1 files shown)


jpsim:
Ouch, this was a heavy one. I'd like to integrate some really rough estimate performance tests, but with tests running on a wide variety of systems, I'm not sure it's possible to do reliably. Do you have any ideas?

Source/SourceKittenFramework/String+SourceKitten.swift
```diff
@@ -101,14 +120,22 @@ extension NSString {
...
+                let element = rangesAndLine[$0]
+                return (line: element.line.index, character: offset - element.range.location + 1)
+            }
+        }
+        
+        func lineAndCharacterForByteOffset(offset: Int) -> (line: Int, character: Int)? {
+            let index = rangesAndLine.indexOf { NSLocationInRange(offset, $0.byteRange) }
+            return index.map {
+                let element = rangesAndLine[$0]
+                return (line: element.line.index, character: offset - element.range.location + 1)
```
norio-nomura: @jpsim: Sorry, I have a bug here. The `character` is not correct. I'll fix that.

norio-nomura: pushed: https://github.com/jpsim/SourceKitten/pull/129

norio-nomura:
Testing on CI? I don't have enough knowledge about that.
I know that Xcode has performance test feature triggering by performance regressions. But that needs setting base performance thresholds on real machine and may not fit CI services.

PR #129, opened 2015-12-25
Fix `NSString.lineAndCharacterForByteOffset(_:)` returns wrong character

norio-nomura:
The character should be UTF16Index distance as I mentioned in https://github.com/realm/SwiftLint/pull/291#discussion_r48430566

### DIFF
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -135,7 +135,18 @@ extension NSString {
             let index = rangesAndLine.indexOf { NSLocationInRange(offset, $0.byteRange) }
             return index.map {
                 let element = rangesAndLine[$0]
-                return (line: element.line.index, character: offset - element.range.location + 1)
+                
+                let character: Int
+                let content = element.line.content
+                let length = offset - element.byteRange.location + 1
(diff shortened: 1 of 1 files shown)


Source/SourceKittenFramework/String+SourceKitten.swift
```diff
@@ -135,7 +135,18 @@ extension NSString {
...
                 let element = rangesAndLine[$0]
-                return (line: element.line.index, character: offset - element.range.location + 1)
+                
+                let character: Int
+                let content = element.line.content
+                let length = offset - element.byteRange.location + 1
+                if length == element.byteRange.length {
+                    character = content.utf16.count
+                } else {
+                    let endIndex = content.utf8.startIndex.advancedBy(length)
```
jpsim: this looks fairly expensive to calculate :grimacing:

norio-nomura: The content is small enough.
Linting KeychainAccess is still under 4 seconds. :smile:
```
swiftlint lint  3.64s user 0.12s system 90% cpu 4.155 total
```

jpsim:
Thanks for finding and addressing this so quickly!

norio-nomura:
:pray:

PR #130, opened 2015-12-25
Optimize `NSString. byteRangeToNSRange(_:)`

norio-nomura:
- Simplify `NSString.CacheContainer`  
Remove `byteOffsetIndexPairs`
- Add `NSString.CacheContainer.locationFromByteOffset(_:)`  
Rewrite `NSString. byteRangeToNSRange(_:)` to use that.

No performance gain by this PR because ` byteRangeToNSRange` is not heavily used yet.
But I think that will be heavily used for fixing such as https://github.com/realm/SwiftLint/issues/294.


jpsim:
Big thanks for this :smile:

norio-nomura:
:smile:

PR #135, opened 2016-01-09
Add `Structure.init(sourceKitResponse:)`

norio-nomura:
SourceKit `editor.open` response can be shared with `SyntaxMap.init(sourceKitResponse:)` for reducing calls to SourceKit.


jpsim:
This makes sense. :+1:

PR #139, opened 2016-01-16
Change method of duplicate string in `NSString.CacheContainer`

norio-nomura:
Fix https://github.com/realm/SwiftLint/issues/379

### DIFF
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -54,8 +54,8 @@ extension NSString {
             //
             // A reference to `NSString` is held by every cast `String` along with their views and
             // indices.
-            let cString = string.cStringUsingEncoding(NSUTF8StringEncoding)
-            let string = String(CString: cString, encoding: NSUTF8StringEncoding)!
+            let data = string.dataUsingEncoding(NSUTF8StringEncoding)
+            let string = String(data: data, usingEncoding: NSUTF8StringEncoding)
             utf8View = string.utf8
 
             var start = 0       // line start
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -54,8 +54,7 @@ extension NSString {
             //
             // A reference to `NSString` is held by every cast `String` along with their views and
             // indices.
-            let data = string.dataUsingEncoding(NSUTF8StringEncoding)
-            let string = String(data: data, usingEncoding: NSUTF8StringEncoding)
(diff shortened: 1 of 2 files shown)


Source/SourceKittenFramework/String+SourceKitten.swift
```diff
@@ -54,8 +54,7 @@ extension NSString {
             //
             // A reference to `NSString` is held by every cast `String` along with their views and
             // indices.
-            let cString = string.cStringUsingEncoding(NSUTF8StringEncoding)
-            let string = String(CString: cString, encoding: NSUTF8StringEncoding)!
+            let string = string.mutableCopy() as! String
```
jpsim: Why is this `mutableCopy()` and not just `copy()`?

jpsim:
Thanks for the fix! I'm just not sure why it has to be a mutable copy.

Source/SourceKittenFramework/String+SourceKitten.swift
norio-nomura: `NSString.copy` returns `self` because that is immutable. `NSString.mutableCopy` returns new instance of `NSString` with copied buffer. On this case, the `string` holds reference of the `CacheContainer` through associated object.
On baa8a3a, `ByteOffsetCache`(renamed to `CacheContainer`) held reference of `string` by [holding `String.UTF8Index`](https://github.com/norio-nomura/SourceKitten/blob/baa8a3a/Source/SourceKittenFramework/String%2BSourceKitten.swift#L34) and that caused circular reference.
On 6656224, it resolved by changing to [create new instance](https://github.com/norio-nomura/SourceKitten/blob/6656224/Source/SourceKittenFramework/String%2BSourceKitten.swift#L51-L52).
Using `mutableCopy()` replaces that.

segiddins: `NSString(string: string) as String` ?

jpsim: Thanks for the detailed explanation!

norio-nomura: @segiddins 
Apple's document said:
> - initWithString:  
> Returns an NSString object initialized by copying the characters from another given string.

But, you can confirm that is incorrect as following:
```swift
import Foundation

let str1 = "Hello, playground" as NSString
let str2 = NSString(string: str1)
let str3 = str1.mutableCopy() as! NSString

let id1 = ObjectIdentifier(str1)
let id2 = ObjectIdentifier(str2)
let id3 = ObjectIdentifier(str3)

id1 == id2 // true
id1 == id3 // false
```

norio-nomura:
:pray:

PR #140, opened 2016-01-19
Use `isEqualTo()` instead of `==` for `XPCRepresentable`

norio-nomura:
depends on https://github.com/jpsim/SwiftXPC/pull/9


jpsim:
Used in #146. Thanks @norio-nomura!

norio-nomura:
Thanks! :pray:

PR #141, opened 2016-01-20
Support Swift Package Manager

jpsim:
First stab at integrating with OSS Swift, using `sourcekitdInProc` and Swift Package Manager.

This work is necessary to eventually have SourceKitten work on Linux. This is pretty early on in the process, but may be of interest to some (maybe @norio-nomura :smile: ).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,8 @@ OUTPUT_PACKAGE=SourceKitten.pkg
 VERSION_STRING=$(shell agvtool what-marketing-version -terse1)
 COMPONENTS_PLIST=Source/sourcekitten/Components.plist
 
+SWIFT_SNAPSHOT=swift-2.2-SNAPSHOT-2016-01-11-a
+
 .PHONY: all bootstrap clean install package test uninstall
 
 all: bootstrap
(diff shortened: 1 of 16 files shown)


jpsim:
I should note that the main blocker in supporting Linux will probably be the lack of SourceKit, since it currently isn't built for the platform (see https://github.com/apple/swift/blob/master/tools/SourceKit/CMakeLists.txt#L21).

norio-nomura:
Wow, great!
As I tested linking to `sourcekitdInProc` several days ago, it seemed there is no benefit from using in-proc library for performance. But now it seems there would be for porting to linux.

By the way, how can we maintain support building on both of SPM and Xcode? Does we need two branches?

Makefile
```diff
@@ -66,3 +68,11 @@ archive:
 	carthage archive SourceKittenFramework
 
 release: package archive
+
+spm_bootstrap:
+	cd /usr/local/include; svn export --force http://llvm.org/svn/llvm-project/cfe/trunk/include/clang-c/
+	cp clang_c_module.modulemap /usr/local/include/clang-c/module.modulemap
+	cp /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/libclang.dylib /usr/local/lib/
```
norio-nomura: swift-2.2-SNAPSHOT contains `libclang.dylib` at `/Library/Developer/Toolchains/swift-2.2-SNAPSHOT-2016-01-11-a.xctoolchain/usr/lib/libclang.dylib`. It would be better using that?

jpsim: Yes, I didn't know that. Thanks!

norio-nomura:
It needs adding `/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/lib` to `FRAMEWORK_SEARCH_PATHS`, `LD_RUNPATH_SEARCH_PATHS` and `LIBRARY_SEARCH_PATHS` in `project.pbxproj`.
I seems we need to switch `Tests/Fixtures` on tests by swift's toolchain, because mangling has been changed on updates of swift. 🤔

jpsim:
FYI this branch is a work in progress and will be heavily rebased & modified, so please don't depend on it being stable for the time being.

jpsim:
This now depends on #146, which is reviewable.

jpsim:
Ok, this is no longer work in progress and only has minimal changes that are specific to SPM. You can review the diff with #146 here: https://github.com/jpsim/SourceKitten/compare/remove-swiftxpc...spm

NachoSoto:
:clap:

norio-nomura:
:clap:

PR #146, opened 2016-01-22
Move away from XPC

jpsim:
This was extracted from #141 and unlike that PR, these commits should be clean-ish and reviewable for those interested /cc @norio-nomura.


NachoSoto:
:clap:

norio-nomura:
:+1:

PR #148, opened 2016-01-27
Replace results of `sourcekitd_uid_get_string_ptr()` with enum's rawValue

norio-nomura:
By applying this, linting Carthage 0.11 by SwiftLint(1fca735):
- Duration is reduced from 22sec to 19sec.
- Persistent Bytes on termination is reduced from 584.61MB to 487.86MB.

Instruments
From:
screenshot 2016-01-27 22 17 12
To:
screenshot 2016-01-27 22 17 26

## Why?
`String(UTF8String:)`'s back end is `NSString`.
On using `String` as key of Dictionary, it uses `hashValue`.
When `String` is `NSString`, `_stdlib_NSStringNFDHashValue` is always called.
https://github.com/apple/swift/blob/swift-2.2-branch/stdlib%2Fpublic%2Fcore%2FString.swift#L484
That uses `NSString.decomposedStringWithCanonicalMapping` for Unicode Normalization Form D.
https://github.com/apple/swift/blob/swift-2.2-branch/stdlib%2Fpublic%2Fstubs%2FSwiftNativeNSXXXBase.mm.gyb#L118
It seems `decomposedStringWithCanonicalMapping` leaks some memory on above process.
Reduced `CFString (mutable)` and `CFString (store)` on Instruments are that maybe.
Those enum's String rawValue is used instead of them.

### DIFF
--- a/Source/SourceKittenFramework/Request.swift
+++ b/Source/SourceKittenFramework/Request.swift
@@ -104,6 +104,9 @@ internal func stringForSourceKitUID(uid: sourcekitd_uid_t) -> String? {
     if let string = uidStringMap[uid] {
         return string
     } else if let uidString = String(UTF8String: sourcekitd_uid_get_string_ptr(uid)) {
+        let uidString = SwiftDocKey(rawValue: uidString)?.rawValue ??
+            SwiftDeclarationKind(rawValue: uidString)?.rawValue ??
+            SyntaxKind(rawValue: uidString)?.rawValue ?? uidString
         uidStringMap[uid] = uidString
         return uidString
     }


jpsim:
Sweet, thanks for diving into this!

norio-nomura:
`SwiftDocKey`, `SwiftDeclarationKind`, `SyntaxKind` do not cover all of uid string from sourcekitd. We can get additional performance improvement by adding them.
Some of them are defined and used in SwiftLint's rules. That will be motivation for defining them in SourceKitten.

jpsim:
Absolutely! We should also have added a comment here explaining the optimization, otherwise it's easy for someone unfamiliar with your findings to assume it's just weird code and simplify :smile:

norio-nomura:
I'll file a PR adding comment for explaining about this PR and adding uid strings.

PR #149, opened 2016-01-28
Use Swift's native `String` instead of `NSString` based `String`

norio-nomura:
- Add `sourceKittenRawValueStringFrom(_:)` and comments for #148
- By applying this,The duration of linting Carthage 0.11 by SwiftLint(8af5c4e)
 with SourceKitten(6025179) is reduced from:
```
swiftlint lint --config ~/.swiftlint-test.yml  18.34s user 1.14s system 83% cpu 23.271 total
```
to:
```
swiftlint lint --config ~/.swiftlint-test.yml  17.60s user 1.13s system 85% cpu 22.014 total
```

Adding enums mentioned in #148 will be separated PR.

### DIFF
--- a/Source/SourceKittenFramework/Request.swift
+++ b/Source/SourceKittenFramework/Request.swift
@@ -104,13 +104,35 @@ internal func stringForSourceKitUID(uid: sourcekitd_uid_t) -> String? {
     if let string = uidStringMap[uid] {
         return string
     } else if let uidString = String(UTF8String: sourcekitd_uid_get_string_ptr(uid)) {
-        let uidString = SwiftDocKey(rawValue: uidString)?.rawValue ??
-            SwiftDeclarationKind(rawValue: uidString)?.rawValue ??
-            SyntaxKind(rawValue: uidString)?.rawValue ?? uidString
+        /*
+        `String` created by `String(UTF8String:)` is based on `NSString`.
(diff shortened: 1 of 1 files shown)


Source/SourceKittenFramework/Request.swift
```diff
@@ -103,16 +104,59 @@ Cache SourceKit requests for strings from UIDs
...
+ 
+- parameter bytes: UnsafePointer<Int8>
+- parameter length: length of bytes
+
+- returns: String Swift's native String
+*/
+private func swiftStringFrom(bytes: UnsafePointer<Int8>, length: Int) -> String? {
+    let pointer = UnsafeMutablePointer<Int8>(bytes)
+    return String(bytesNoCopy: pointer, length: length, encoding: NSUTF8StringEncoding,
+        freeWhenDone: false).map { "\($0)" }
```
jpsim: Why the `map`? It just goes from `String?` to `String?`.

neilpa: It's converting from the `NSString` backed swift string to a native swift string. The c'tor is actually `-[NSString initWithBytesNoCopy:length:encoding:freeWhenDone]`

jpsim: Awesome.

jpsim:
Thanks for this!

norio-nomura:
:pray:

PR #152, opened 2016-01-30
Change `Line` from tuple to struct with extra properties `range` and `byteRange`

norio-nomura:
This allows more performance optimizations on SwiftLint.


jpsim:
Cool. :+1: Thanks

norio-nomura:
:pray:

Issue #154, opened 2016-01-31
Run unit tests with SPM


jpsim:
The unit tests are only run in the Xcode job at the moment, it'd be good to also run the unit tests with SPM.

jpsim:
Done in #158.

PR #158, opened 2016-02-03
[WIP][SPM] Add `make spm_test`

norio-nomura:
#154

Notes:
- `swift-DEVELOPMENT-SNAPSHOT-2016-01-25-a` contains `libswiftXCTest.dylib`. But that can not be used on building by SPM because that does not contain module. So, this PR uses `swift-corelibs-xctest` as dependency.
- Use forked [`norio-nomura/swift-corelibs-xctest`](https://github.com/norio-nomura/swift-corelibs-xctest) as dependency because original repository has no version tags yet.
- Test cases are needed conforming to `XCTestCaseProvider`.
- Add `import Foundation` explicitly.

**Test fails currently.**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
   include:
     - script: make test
       env: JOB=Xcode
-    - script: make spm
+    - script: make spm_test
       env: JOB=SPM
       before_install:
         - make swift_snapshot_install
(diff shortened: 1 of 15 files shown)


jpsim:
This is great @norio-nomura! I want to get this in, even with the failing tests, so I'll comment those out for now. It's valuable enough to test the non-failing stuff on SPM.

:clap:

norio-nomura:
Thanks! :pray:

PR #159, opened 2016-02-03
Typed swift statements

garnett:


jpsim:
:+1: thanks for this @garnett!

PR #163, opened 2016-02-05
Use `sourcekitd_response_dispose()`

norio-nomura:
Instruments on linting Carthage 0.12,
maximum Persistent Bytes is reduced from 508.49MB to 386.19MB.
From:
screenshot 2016-02-05 19 21 17
To:
screenshot 2016-02-05 19 22 15
Total number of allocation does not increase. (If disposed memory was reused by sourcekitd, it increases.)

Running Time of calling `sourcekitd_response_dispose()` is 282ms:
screenshot 2016-02-05 19 41 43

### DIFF
--- a/Source/SourceKittenFramework/Request.swift
+++ b/Source/SourceKittenFramework/Request.swift
@@ -253,7 +253,9 @@ public enum Request {
         dispatch_once(&sourceKitInitializationToken) {
             sourcekitd_initialize()
         }
-        return fromSourceKit(sourcekitd_response_get_value(sourcekitd_send_request_sync(sourcekitObject))) as! [String: SourceKitRepresentable]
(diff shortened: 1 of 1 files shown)


jpsim:
:+1: this will be even more important for apps that use SourceKitten for longer periods of time than just CLI tools. Thanks!

norio-nomura:
Thanks! :pray:

I thought we should also use `sourcekitd_request_release()` after calling `sourcekitd_send_request_sync` in `send()`, and tested.
It was ok for swiftlint, but not for SourceKittenFrameworkTest on calling from `sendCursorInfoRequest()`.
Sorry, I have not yet read codes of SourceKitten other than related to SwiftLint, so I couldn't include that.

PR #165, opened 2016-02-06
Fix crash on offset points end of string

### LINKED ISSUE #164: fatal error on linting `09385-swift-structtype-get.swift` with SwiftLint
SwiftLint crashes on linting [09385-swift-structtype-get.swift](https://github.com/apple/swift/blob/master/validation-test/compiler_crashers_fixed/09385-swift-structtype-get.swift)

```sh
% curl --silent https://raw.githubusercontent.com/apple/swift/master/validation-test/compiler_crashers_fixed/09385-swift-structtype-get.swift | swiftlint lint --use-stdin
Loading configuration from '.swiftlint.yml'
fatal error: : file /Users/norio/Documents/workspace/github/SwiftLint/Carthage/Checkouts/SourceKitten/Source/SourceKittenFramework/String+SourceKitten.swift, line 116
[1]    66323 done                          curl --silent  | 
       66324 illegal hardware instruction  swiftlint lint --use-stdin
```

It caused by that sometime sourcekitd returns "key.nameoffset"+ "key.namelength" == `utf8.count`

norio-nomura:
- `locationFromByteOffset` crashed on `byteOffset` points end of string  
`byteOffset` may be out of bounds when sourcekitd points end of string.  
Fix #164
- `byteOffsetFromLocation` crashed on `location` points end of string  
`location` may be out of bounds when `NSRegularExpression` points end of string.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+## Master
+
+##### Breaking
+
+* None.
+
+##### Enhancements
+
+* None.
+
+##### Bug Fixes
+
+* Fix crash when offset points end of string.  
+  [Norio Nomura](https://github.com/norio-nomura)
+  [#164](https://github.com/realm/SwiftLint/issues/379)
+
 ## 0.9.0
 
 ##### Breaking
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -112,10 +112,11 @@ extension NSString {
             if lines.isEmpty {
                 return 0
             }
-            guard let index = lines.indexOf({ NSLocationInRange(byteOffset, $0.byteRange) }) else {
+            let index = lines.indexOf({ NSLocationInRange(byteOffset, $0.byteRange) })
+            // byteOffset may be out of bounds when sourcekitd points end of string.
+            guard let line = (index.map { lines[$0] } ?? lines.last) else {
                 fatalError()
             }
-            let line = lines[index]
             let diff = byteOffset - line.byteRange.location
             if diff == 0 {
                 return line.range.location
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -140,10 +140,11 @@ extension NSString {
             if lines.isEmpty {
                 return 0
             }
-            guard let index = lines.indexOf({ NSLocationInRange(location, $0.range) }) else {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+
+##### Enhancements
+
+* None.
+
+##### Bug Fixes
+
+* Fix crash when offset points end of string.  
+  [Norio Nomura](https://github.com/norio-nomura)
+  [#164](https://github.com/realm/SwiftLint/issues/379)
```
jpsim: url is mismatched

jpsim:
This makes me wonder if we're misusing the SourceKit API. I'm not sure if @akyrtzi or @nkcsgexi are allowed to comment here, but they would know for sure.

Regardless, this fixes real crashes and doesn't appear to introduce regressions since the only path affected is one that would previously crash, so let's get this in. Thanks @norio-nomura!

norio-nomura:
Sorry, this issue had been introduced by my optimizations. :bow:
The crash did not happen on my previous testing https://github.com/realm/SwiftLint/issues/380, because no rules passed such parameters to those functions that cause crashes.

CHANGELOG.md
norio-nomura: Oops!

akyrtzi:
@jpsim, not sure what the issue is here, but if you think there's an issue with ranges returned by sourcekitd I'd be interested in looking into a test case. If you can provide a json dictionary request or series of requests that can be passed to the sourcekitd-repl tool, along with accompanying files, that would be ideal for me.

norio-nomura:
Sorry, my bad.
My expression "out of bounds" was incorrect and made @jpsim misunderstood.
I should write "~offset points the next of end of string".
@akyrtzi I think sourcekitd behaves correctly in that case. issue was in my code.

I didn't properly understand that when `NSRage` representing the range of end in string, `NSMaxRange()` returns next offset of end of string.
Example on `NSRegularExpression`:
```swift
import Foundation

var str = "Hello, playground"

str.utf16.count // 17

let regex = try! NSRegularExpression(pattern: "nd", options: [])
let range = NSRange(location: 0, length: str.utf16.count)

// NSRange representation of end of string
let matchRange = regex.matchesInString(str, options: [], range: range).first!.range

let offset = NSMaxRange(matchRange)   // 17
NSLocationInRange(offset, matchRange) // false, because offset points next of end of string.

let indexOfOffset = str.utf16.startIndex.advancedBy(offset)
str.utf16[indexOfOffset] // error!  offset is invalid.
```
sourcekitd's offset and length are behaving in same manner as above.

:bow:

PR #168, opened 2016-02-08
revert to previous approach for SPM support, requiring no llbuild flags

jpsim:
Although it's destructive regarding files on the user's system, this approach allows `swift build` without arguments to succeed. /cc @norio-nomura

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,17 +16,8 @@ OUTPUT_PACKAGE=SourceKitten.pkg
 VERSION_STRING=$(shell agvtool what-marketing-version -terse1)
 COMPONENTS_PLIST=Source/sourcekitten/Components.plist
 
-SWIFT_SNAPSHOT=swift-DEVELOPMENT-SNAPSHOT-2016-01-25-a
-
-SPM=/Library/Developer/Toolchains/$(SWIFT_SNAPSHOT).xctoolchain/usr/bin/swift build
-SPM_INCLUDE=/Library/Developer/Toolchains/$(SWIFT_SNAPSHOT).xctoolchain/usr/local/include
-SPM_LIB=/Library/Developer/Toolchains/$(SWIFT_SNAPSHOT).xctoolchain/usr/lib
-# for including "clang-c"
-SPMFLAGS=-Xcc -ISource/Clang_C -Xcc -I$(SPM_INCLUDE)
-# for linking sourcekitd and clang-c
-SPMFLAGS+= -Xcc -F$(SPM_LIB) -Xlinker -F$(SPM_LIB) -Xlinker -L$(SPM_LIB)
-# for loading sourcekitd and clang-c
-SPMFLAGS+= -Xlinker -rpath -Xlinker $(SPM_LIB)
+SWIFT_SNAPSHOT=swift-DEVELOPMENT-SNAPSHOT-2016-02-03-a
+SWIFT_BUILD_COMMAND=/Library/Developer/Toolchains/$(SWIFT_SNAPSHOT).xctoolchain/usr/bin/swift build
 
 .PHONY: all bootstrap clean install package test uninstall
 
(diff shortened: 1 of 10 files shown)


norio-nomura:
:+1:

Makefile
```diff
@@ -83,16 +74,26 @@ swift_snapshot_install:
...
+spm_bootstrap: spm_teardown
+	ln -s /Library/Developer/Toolchains/$(SWIFT_SNAPSHOT).xctoolchain/usr/local/include/clang-c /usr/local/include/clang-c
+	sudo cp Source/Clang_C/module.modulemap /usr/local/include/clang-c/module.modulemap
+	ln -s /Library/Developer/Toolchains/$(SWIFT_SNAPSHOT).xctoolchain/usr/lib/libclang.dylib /usr/local/lib/libclang.dylib
+	mkdir -p /usr/local/include/sourcekitdInProc
+	ln -s /Library/Developer/Toolchains/$(SWIFT_SNAPSHOT).xctoolchain/usr/lib/sourcekitd.framework/Headers/sourcekitd.h /usr/local/include/sourcekitdInProc/sourcekitd.h
+	curl https://static.realm.io/libsourcekitdInProc/$(SWIFT_SNAPSHOT)/libsourcekitdInProc.dylib -o /usr/local/lib/libsourcekitdInProc.dylib
+
+spm_teardown:
+	rm -rf /usr/local/include/clang-c /usr/local/lib/libclang.dylib
```
norio-nomura: `spm_teardown` does not remove `/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2016-02-03-a.xctoolchain/usr/local/include/clang-c/module.modulemap`

jpsim: Since the files in the toolchain don't change, I think I'll avoid the use of `sudo` here and copy the files instead. Then `spm_teardown` will remove the module map when deleting `clang-c`.

jpsim:
@norio-nomura what do you think we should do to further improve SPM support?

I'm hesitant to move the `sourcekitten` executable out of this repo, but that seems to be the only way to avoid the build-delete-build approach currently taken by SwiftLint's Makefile.

jpsim:
> I'm hesitant to move the sourcekitten executable out of this repo, but that seems to be the only way to avoid the build-delete-build approach currently taken by SwiftLint's Makefile.

Actually, even that's not enough, since we'll still have the tests. SPM doesn't have an official way to run tests yet, so we'd need to move that out of the repo too! Maybe we just wait and put pressure on [SR-688](https://bugs.swift.org/browse/SR-688) to fix this in SPM since we can't really fix this on our end.

In the meantime, I think I'll make a `SourceKittenSPM` repo that mirrors releases from this one, but without the `sourcekitten` and `SourceKittenFrameworkTests` directories.

Package.swift
```diff
@@ -10,8 +10,9 @@ let package = Package(
       dependencies: [.Target(name: "SourceKittenFramework")]),
   ],
   dependencies: [
-    .Package(url: "https://github.com/drmohundro/SWXMLHash.git", majorVersion: 2),
-    .Package(url: "https://github.com/Carthage/Commandant.git", majorVersion: 0, minor: 8),
-    .Package(url: "https://github.com/norio-nomura/swift-corelibs-xctest.git", majorVersion: 0),
+    .Package(url: "https://github.com/jpsim/SourceKit", majorVersion: 1),
+    .Package(url: "https://github.com/drmohundro/SWXMLHash", majorVersion: 2),
+    .Package(url: "https://github.com/Carthage/Commandant", majorVersion: 0, minor: 8),
+    .Package(url: "https://github.com/norio-nomura/swift-corelibs-xctest", majorVersion: 0),
```
norio-nomura: On `make spm_test` in SwiftLint:
> Cloning https://github.com/norio-nomura/swift-corelibs-xctest
> Using version 0.1.0-test of package swift-corelibs-xctest
> Cloning https://github.com/behrang/YamlSwift.git
> Using version 1.4.2 of package YamlSwift
> Cloning https://github.com/norio-nomura/swift-corelibs-xctest.git
> /usr/local/bin/git clone --recursive --depth 10 https://github.com/norio-nomura/swift-corelibs-xctest.git /Users/norio/Documents/workspace/github/SwiftLint/Packages/swift-corelibs-xctest
> fatal: destination path '/Users/norio/Documents/workspace/github/SwiftLint/Packages/swift-corelibs-xctest' already exists and is not an empty directory.
>
> error: Failed to clone https://github.com/norio-nomura/swift-corelibs-xctest.git to /Users/norio/Documents/workspace/github/SwiftLint/Packages/swift-corelibs-xctest

It seems we need making *exactly same urls* in each `Package.swift`, can not mix with/without `.git` on same repo. :confused:

jpsim:
Ok, I'll put the `.git` back in there

jpsim:
Re-added the `.git`s in 3f773f5.

norio-nomura:
Thanks! :pray:

norio-nomura:
> I'm hesitant to move the sourcekitten executable out of this repo, but that seems to be the only way to avoid the build-delete-build approach currently taken by SwiftLint's Makefile.

I pushed PR for testing https://github.com/realm/SwiftLint/pull/516 without build-delete-build.

norio-nomura:
:+1: https://travis-ci.org/realm/SwiftLint/jobs/107926696

PR #170, opened 2016-02-09
move spm_bootstrap and spm_teardown to dedicated scripts

jpsim:
also explicitly add antitypical/Result dependency to `Package.swift` due to https://github.com/realm/SwiftLint/pull/516#issuecomment-181734495.

### DIFF
--- a/script/spm_bootstrap
+++ b/script/spm_bootstrap
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+SWIFT_SNAPSHOT=$1
+
+cp -r /Library/Developer/Toolchains/$SWIFT_SNAPSHOT.xctoolchain/usr/local/include/clang-c /usr/local/include/clang-c
+cp Source/Clang_C/module.modulemap /usr/local/include/clang-c/module.modulemap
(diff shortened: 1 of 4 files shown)


script/spm_bootstrap
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+SWIFT_SNAPSHOT=$1
+
+cp -r /Library/Developer/Toolchains/$SWIFT_SNAPSHOT.xctoolchain/usr/local/include/clang-c /usr/local/include/clang-c
+cp Source/Clang_C/module.modulemap /usr/local/include/clang-c/module.modulemap
```
norio-nomura: For using from `curl …| sh`, `module.modulemap` also needed to create here or download from.

jpsim: right, thanks

jpsim:
Now SwiftLint's Makefile can look like this:

```
spm_bootstrap: spm_teardown
  curl https://raw.githubusercontent.com/jpsim/SourceKitten/master/script/spm_bootstrap | bash -s $(SWIFT_SNAPSHOT)

spm_teardown:
  curl https://raw.githubusercontent.com/jpsim/SourceKitten/master/script/spm_teardown | bash
```

Issue #167, opened 2016-02-06
Use toolchain from xcode-select


jpsim:

norio-nomura:
As I saw https://github.com/jpsim/SourceKitten/pull/153#issuecomment-180985100, I feel we need doing this.

norio-nomura:
did I misunderstand this?
I understood this as:
- `sourcekitten` loads `sourcekitd.framework` from directory tree that `xcode-select -p` pointing.  
- `xcode-select -p` are referred on runtime.  

I wrote #172 and #176 based on that…

jpsim:
Yes, you understood this correctly and #172 & #176 seem like great solutions to this! I'm just trying to understand all the complexities required by the implementations in those PRs, and I'm starting to understand.

Looks like you did great work there! It's just taking me a long time to comprehend what all of it means.

norio-nomura:
Thanks, I was relieved. :relieved:

jpsim:
Done by @norio-nomura 👏 🎉

PR #172, opened 2016-02-12
[WIP] Use `xcode-select -p` (weak link approach)

norio-nomura:
#167 
#### Overview:

- `libclang.dylib` and `sourcekitd.framework` are weakly linked to `SourceKittenFramework`.
- toolchain pathes has been removed from `LC_RPATH` on `SourceKittenFramework`.
- `sourcekitten` loads `libclang.dylib` and `sourcekitd.framework` from active developer directory by calling `checkDepndenciesOfSourceKittenFramework()` in `main.swift`
- `checkDepndenciesOfSourceKittenFramework()` calls `execv` for re-launching main executable for loading `libclang.dylib` and `sourcekitd.framework`. So, it might not be used on some dependent client of `SourceKittenFramework`

#### `checkDepndenciesOfSourceKittenFramework()` does following:

- Check whether `libclang.dylib` and `sourcekitd.framework` are loaded or not
- If not, coordinates `DYLD_*` environment variables and calls `execv`
- New `sourcekitten` process launched by `execv` loads `libclang.dylib` and `sourcekitd.framework` using coordinated `DYLD_*` environment.

#### `DYLD_*` environment variables are coordinated to searching them in following order:
1. `$TOOLCHAIN_DIR/usr/lib` # for debugging on Xcode
2. \`xcode-select -p\` `/Toolchains/XcodeDefault.xctoolchain/usr/lib`
3. `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib`
4. `/Applications/Xcode-beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib`
5. `$HOME/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib`
6. `$HOME/Applications/Xcode-beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib`

3~6 are fallbacks on `xcode-select -p` pointing Command Line Tools OS X for Xcode, because that does not contains `sourcekitd.framework`.

#### Problems
- SPM does not support weak link yet
- `execv` needed. I don't know how large this affects to other clients.

### DIFF
--- a/Source/SourceKittenFramework/DyldHelper.swift
+++ b/Source/SourceKittenFramework/DyldHelper.swift
@@ -1,0 +1,192 @@
+//
+//  DyldHelper.swift
+//  sourcekitten
+//
+//  Created by Norio Nomura on 2/11/16.
+//  Copyright © 2016 SourceKitten. All rights reserved.
+//
+
+import Foundation
+
+/// Environment Variable that indicates current process is launched through `execv()`
+private let sourceKittenFrameworkLoadingkey = "SOURCEKITTEN_FRAMEWORK_LOADING"
+
+private let libclangPath = "libclang.dylib"
+private let sourcekitdPath = "sourcekitd.framework/Versions/A/sourcekitd"
+
+/// Checks whether depending dynamic link libraries are loaded or not.
+/// If not, coordinates `DYLD_*` environment variables, calls `execv()` and never returns.
+public func checkDepndenciesOfSourceKittenFramework() {
+    if libraryIsLoaded(libclangPath) && libraryIsLoaded(sourcekitdPath) {
+        return
+    }
+
+    // following codes will be executed when dependencies will not be loaded.
+    // On sourcekitten installation, LC_RPATH of SourceKittenFramework does not contain pathes
+    // which can find libclang.dylib and sourcekitd.framework.
+
+    coordinateEnvironmentVariablesForDYLD()
+    setenv(sourceKittenFrameworkLoadingkey, "1", 1)
+
+    let arguments = NSProcessInfo.processInfo().arguments
+    let path = arguments[0]
+    let argv = CStringArray(arguments)
+    execv(path, argv.pointers) // never returns on normal flow
+    let error = String(UTF8String: strerror(errno))! // swiftlint:disable:this force_unwrapping
+    fatalError("execv failed: \(error)")
+}
+
+/// Returns true if library is not loaded.
+/// - Parameter path: string of library's path
+private func libraryIsLoaded(path: String) -> Bool {
+    let library = dlopen(path, RTLD_LAZY)
+    if library != nil {
+        dlclose(library)
+        return true
+    } else  if let _ = NSProcessInfo.processInfo().environment
+        .indexForKey(sourceKittenFrameworkLoadingkey) {
+        fatalError("Can't load \(path)")
+    }
+    return false
+}
+
+/// Coordinates "DYLD_LIBRARY_PATH" and "DYLD_FRAMEWORK_PATH" for `execv`ed process.
+private func coordinateEnvironmentVariablesForDYLD() {
+    let fileManager = NSFileManager.defaultManager()
+    let developerDirXcode = "/Applications/Xcode.app/Contents/Developer"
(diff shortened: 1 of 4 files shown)


segiddins:
This should also consider `DEVELOPER_DIR`

norio-nomura:
> This should also consider `DEVELOPER_DIR`

As far as I tested, `DEVELOPER_DIR` is reflected to `TOOLCHAIN_DIR` on Xcode.
Ah, but when we run `DEVELOPER_DIR="/Applications/Xcode-beta.app" make` and swiftlint is directly launched by `make`, we should consider that.
I'll add it.

norio-nomura:
> This should also consider `DEVELOPER_DIR`

`DEVELOPER_DIR` environment variable affects to result of `xcode-select -p`.
So, using result of `xcode-select -p` means considering `DEVELOPER_DIR`.
I won't add checking `DEVELOPER_DIR` individually.

Source/SourceKittenFramework/DyldHelper.swift
```diff
@@ -0,0 +1,192 @@
...
+
+    coordinateEnvironmentVariablesForDYLD()
+    setenv(sourceKittenFrameworkLoadingkey, "1", 1)
+
+    let arguments = NSProcessInfo.processInfo().arguments
+    let path = arguments[0]
+    let argv = CStringArray(arguments)
+    execv(path, argv.pointers) // never returns on normal flow
+    let error = String(UTF8String: strerror(errno))! // swiftlint:disable:this force_unwrapping
+    fatalError("execv failed: \(error)")
```
zwaldowski: Why not the stdlib's `Process`?

zwaldowski: If not, it might be better to follow the way the [stdlib does](https://github.com/apple/swift/blob/master/stdlib/private/SwiftPrivate/SwiftPrivate.swift#L75-L97) instead of CStringArray.

Source/SourceKittenFramework/DyldHelper.swift
```diff
@@ -0,0 +1,192 @@
...
+            if let fullPath = path?.stringByAppendingPathComponent("/usr/lib")
+                where fileManager.fileExistsAtPath(fullPath) {
+                    return fullPath
+            }
+            return nil
+        }
+
+    #if SWIFT_PACKAGE
+        pathes = ["/Library/Developer/Toolchains/swift-latest"] + pathes
+    #endif
```
zwaldowski: `/Applications/` and `/Library/Developer/` are not guaranteed constants by the platform; Foundation's search path APIs should be used instead. Additionally, Swift toolchains can be installed to `~/Library/Developer`.

zwaldowski:
And also consider the `XCODE_DEFAULT_TOOLCHAIN_OVERRIDE` env variable.

I don't think relaunching automatically is the right long-term solution for clients that link or embed SourceKittenFramework directly — for instance, that approach doesn't work right for a Cocoa app embedding SourceKittenFramework. We should `dlsym` the symbols we're using after finding the toolchain; this would also allow us to sandbox toolchain lookup into a single type, and would force us to harden the leaking of any stuff from the SourceKit or libclang headers. I've had success with this on a branch.

Source/SourceKittenFramework/DyldHelper.swift
norio-nomura: Oh, I didn't know `Process.unsafeArgv`.
Thanks!

Source/SourceKittenFramework/DyldHelper.swift
norio-nomura: I should not include that because this weak link approach does not work with SPM.

norio-nomura:
> I don't think relaunching automatically is the right long-term solution

I agree.

norio-nomura:
> And also consider the `XCODE_DEFAULT_TOOLCHAIN_OVERRIDE` env variable.

It seems `XCODE_DEFAULT_TOOLCHAIN_OVERRIDE` does not affect to `TOOLCHAIN_DIR`.
We need that.
Thanks!

norio-nomura:
This was replaced by #176 that using  `dlsym()` approach.

PR #173, opened 2016-02-12
Use `init(minimumCapacity:)` for Dictionary in `fromSourceKit(_:)`

norio-nomura:
By applying this, the duration of `fromSourceKit(_:)` on linting Carthage 0.13 by SwiftLint is reduced from 2490ms to 1536ms.
from:
screenshot 2016-02-11 13 25 18
to:
screenshot 2016-02-11 13 26 12

Why is capacity 9?
The count of dictionaries are following on above condition:

|count|happen|
|----:|-----:|
|    1|  5043|
|    3|274726|
|    4|     2|
|    5|  1216|
|    6|  1359|
|    7| 28351|
|    8| 38175|
|    9| 20653|
|   10|  4403|
|   11|  2513|
|   12|   255|
|   13|   660|
|   14|   173|

This approach couldn't apply before introducing https://github.com/realm/SwiftLint/pull/487, because of increasing persistent memory.

### DIFF
--- a/Source/SourceKittenFramework/Request.swift
+++ b/Source/SourceKittenFramework/Request.swift
@@ -64,7 +64,12 @@ private func fromSourceKit(sourcekitObject: sourcekitd_variant_t) -> SourceKitRe
         }
         return array
     case SOURCEKITD_VARIANT_TYPE_DICTIONARY:
-        var dict = [String: SourceKitRepresentable](minimumCapacity: 9)
+        var count: Int = 0
+        sourcekitd_variant_dictionary_apply(sourcekitObject) { _, _ in
+            count += 1
+            return true
+        }
+        var dict = [String: SourceKitRepresentable](minimumCapacity: count)
         sourcekitd_variant_dictionary_apply(sourcekitObject) { key, value in
             if let key = stringForSourceKitUID(key), value = fromSourceKit(value) {
                 dict[key] = value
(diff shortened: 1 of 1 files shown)


NachoSoto:
Nice :+1: 
I would leave a comment explaining that 9 was picked heuristically by looking at sample data or something like that.

jpsim:
I wonder how much those numbers are representative of all projects versus just Carthage and its dependencies...

Also, dictionaries with a capacity of 3 are 13x as likely to happen as ones with 9, which means we'd allocate 3x as much memory for those as is actually necessary.

I think it'd be worthwhile to expose a SourceKit equivalent to `xpc_dictionary_get_count`, since then we could always just allocate the appropriate size of dictionary when converting.

norio-nomura:
> I think it'd be worthwhile to expose a SourceKit equivalent to xpc_dictionary_get_count, since then we could always just allocate the appropriate size of dictionary when converting.

I have looked into source code of SourceKit for that. But I can't find.
I know this PR's approach would be unacceptable.

jpsim:
> I have looked into source code of SourceKit for that. But I can't find.

That's what I'm saying. SourceKit *doesn't* currently expose it, but it *could* and we could make a PR towards apple/swift that exposes it, since it's available on the XPC side, which SourceKit uses as the representation of its variant objects.

norio-nomura:
I changed using the magic number 9 to counting elements of dictionary.

I tested two variant:
A. Store (key, value) to Array before creating Dictionary.
B. Count elements before creating Dictionary.

Those duration of `fromSourceKit(_:)` on linting Carthage 0.13 by SwiftLint were:
A. 2049ms
B. 1881ms
screenshot 2016-02-13 11 45 42

This may be better than using the magic number 9.

jpsim:
I feel more comfortable with this than the magic number. Thanks :+1:

norio-nomura:
:pray:

PR #176, opened 2016-02-20
Use `xcode-select -p` (dlsym() approach)

norio-nomura:
Another approach for fixing #167

Improved from #172:
- No `execv`
- SPM compatible
- Searching toolchain path is extended by applying reviews on #172.

Progress:
- [x] sourcekitd.framework
- [x] libclang.dylib

PR #178, opened 2016-02-26
Runtime linking of Clang & SourceKit, codegen & update to latest Swift snapshot

### LINKED ISSUE #177: Support Swift Package Manager Testing
`swift-DEVELOPMENT-SNAPSHOT-2016-02-25-a` has come that supports `swift test`.
After SourceKitten support that, new version tag would be needed for SPM dependent.

#176 needs modifying `Package.swift`. So, I think we should support `swift test` before doing #176.

norio-nomura:
Depends on #176
Resolves #177


norio-nomura:
I can't reproduce the crash locally that happens on Travis-CI. :cry:

norio-nomura:
Fixed crash on Travis-CI.

jpsim:
I *really* don't understand why all of this is necessary. Swift can use `.h` headers to resolve symbols in binaries, why do we need to do all this codegen to generate Swift interface source code when `swiftc` can do that from `.h` files at compile time?

norio-nomura:
@jpsim Do you mean writing universal wrapper functions in Assembler or C?

jpsim:
I mean why do we need to generate the Swift interfaces in `interface_gen` at all? Why can't we just continue to use the `.h` interfaces like we have all along?

norio-nomura:
We convert Swift interfaces into wrapper functions. e.g. https://github.com/jpsim/SourceKitten/pull/176/files#diff-0ff653237c4124ffaf3365124e0aaa24R50
```swift
public func sourcekitd_initialize() {
    return _sourcekitd_initialize()
}
private let _sourcekitd_initialize: @convention(c) () -> () = library.loadSymbol("sourcekitd_initialize")
```
When called, that resolves actual symbols from dynamic link library and doing actual function calls.
Sorry, my [documentation](https://github.com/jpsim/SourceKitten/blob/nn-use-xcode-select-p-dlsym/Source%2FSourceKittenFramework%2Flibrary_wrapper%2FREADME.md) is bad.

norio-nomura:
I can't find the way how to catch all function calls without writing wrapper functions for each functions that exported by dynamic link library.

jpsim:
Thanks for those explanations @norio-nomura, I'm beginning to understand. I'm still very curious to see if there's a way to do this without all the code generation. This PR adds 10,570 new lines!

norio-nomura:
Those new lines are library wrappers. If we didn't support SPM, we could generate them on every compile by Xcode.

jpsim:
I understand that, but I'm still curious to find ways to avoid this complexity. It might not be possible.

jpsim:
The more I review this, the more I understand the problem it's solving and the way it's solving it. The more my head hurts too :stuck_out_tongue:.

I have some ideas to simplify the process...

jpsim:
This is really, really impressive work, by the way! :clap:

norio-nomura:
Thanks!

> I have some ideas to simplify the process...

The process is complicated. I'm not good at writing those scripts…

jpsim:
I've added a different approach in #185.

Source/SourceKittenFramework/library_wrapper.swift
```diff
@@ -0,0 +1,145 @@
...
+    })
+
+struct Loader {
+    let paths: [String]
+
+    func load(path: String) -> DynamicLinkLibrary {
+        let fullPaths = paths.map { $0.stringByAppendingPathComponent(path) }.filter { $0.isFile }
+
+        // try all fullPaths that contains target file
+        for fullPath in fullPaths {
```
jpsim: Minor, but could do `for fullPath in fullPaths + [path] {` instead of having that duplicated code a little further down for just `path`.

Source/SourceKittenFramework/library_wrapper.swift
```diff
@@ -0,0 +1,145 @@
...
+            let handle = dlopen(fullPath, RTLD_LAZY)
+            if handle != nil {
+                return DynamicLinkLibrary(path: path, handle: handle)
```
jpsim: Why is this `path: path` and not `path: fullPath`

Source/SourceKittenFramework/library_wrapper.swift
norio-nomura: `paths` contains search paths of loader.
Maybe I should rename that property to `searchPaths`.

jpsim: I think your response is to my other comment, in which case is fine. What I mean here is this:

```swift
// try all fullPaths that contains target file
for fullPath in fullPaths {
    let handle = dlopen(fullPath, RTLD_LAZY)
    if handle != nil {
        return DynamicLinkLibrary(path: path, handle: handle)
    }
}

// try loading with simple path that depends resolving to DYLD
let handle = dlopen(path, RTLD_LAZY)
if handle != nil {
    return DynamicLinkLibrary(path: path, handle: handle)
}
```

could be simplified to this:

```swift
// try all fullPaths that contains target file
// and then again with simple path that depends resolving to DYLD
for fullPath in fullPaths + [path] {
    let handle = dlopen(fullPath, RTLD_LAZY)
    if handle != nil {
        return DynamicLinkLibrary(path: path, handle: handle)
    }
}
```

Source/SourceKittenFramework/library_wrapper.swift
norio-nomura: That is used as:
```swift
private let library = toolchainLoader.load("sourcekitd.framework/Versions/A/sourcekitd")
private let library = toolchainLoader.load("libclang.dylib")
```
On those usage, path is identifier of dynamic link library.
e.g.
```sh
% otool -L /usr/local/Frameworks/SourceKittenFramework.framework/SourceKittenFramework
/usr/local/Frameworks/SourceKittenFramework.framework/SourceKittenFramework:
	/usr/local/opt/sourcekitten/Frameworks/SourceKittenFramework.framework/Versions/A/SourceKittenFramework (compatibility version 1.0.0, current version 1.0.0)
	@rpath/libclang.dylib (compatibility version 1.0.0, current version 700.1.81)
	@rpath/sourcekitd.framework/Versions/A/sourcekitd (compatibility version 0.0.0, current version 700.1.81)
	@rpath/SWXMLHash.framework/Versions/A/SWXMLHash (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1256.14.0)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1256.1.0)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
	@rpath/libswiftCore.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libswiftCoreGraphics.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libswiftDarwin.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libswiftDispatch.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libswiftFoundation.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libswiftObjectiveC.dylib (compatibility version 0.0.0, current version 0.0.0)
```
`@rpath/libclang.dylib` and `@rpath/sourcekitd.framework/Versions/A/sourcekitd` direct to DYLD as:
"search `libclang.dylib` from LC_RPATH"
"search `sourcekitd.framework/Versions/A/sourcekitd` from LC_RPATH"
I followed those DYLD's manner.

jpsim: :+1:

Source/SourceKittenFramework/library_wrapper.swift
norio-nomura: `path` is library identifier and not searchPath. `fullPaths` already joined `searchPath`+`path` on above line as:
```swift
let fullPaths = paths.map { $0.stringByAppendingPathComponent(path) }.filter { $0.isFile }
```

jpsim: I don't think you're reading my suggestion correctly. My changes do the same things as before, just more compactly :wink:.

norio-nomura: Oh! I under stand🕴 Thanks! :pray:

norio-nomura:
Updated!

Source/SourceKittenFramework/library_wrapper.swift
```diff
@@ -0,0 +1,140 @@
...
+        return nil
+    }
+
+    let task = NSTask()
+    task.launchPath = pathOfXcodeSelect
+    task.arguments = ["-p"]
+
+    let pipe = NSPipe()
+    task.standardOutput = pipe
+    task.standardError = pipe
```
jpsim: why are we piping standard error at all? this could mangle the output.

jpsim:
I think we can remove all the `allTests` now that XCTest from SPM can automatically pick them up. I think it can, anyway.

jpsim:
Otherwise, this looks great and ready to merge! Really impressive work, :clap: :+1:

Source/SourceKittenFramework/library_wrapper.swift
norio-nomura: :+1:

norio-nomura:
> I think we can remove all the allTests now that XCTest from SPM can automatically pick them up.

I didn't know that. I'll remove.

jpsim:
> I didn't know that. I'll remove.

I'm not 100% sure about that, you'll have to confirm.

norio-nomura:
I confirmed that removing `XCTestCaseProvider.allTests` works.
On build log, https://travis-ci.org/jpsim/SourceKitten/jobs/116659522#L1320 comment outed `SwiftDocsTests testBicycle` on `allTests` had been already tested.
As reading [README.md of swift-corelibs-xctest](https://github.com/apple/swift-corelibs-xctest/blob/master/README.md#additional-considerations-for-swift-on-linux), `var allTests` has been changed to `static var allTests` for Linux.
We should remove obsoleted `var allTests`.

norio-nomura:
Updated :zzz:

jpsim:
:+1: should we merge and cut a release?

.travis.yml
```diff
@@ -13,8 +13,8 @@ matrix:
     - script: make spm_test
       env: JOB=SPM
       before_install:
+        - make bootstrap
```
jpsim: is this still necessary?

norio-nomura: There is a test that depends on existence of `Carthage/Checkouts/Commandant`.
--
Norio Nomura @norio_nomura

2016/03/18 0:51$B!"(BJP Simard <notifications@github.com> $B$N%a%C%;!<%8(B:

jpsim: oh of course :+1:

norio-nomura:
Thanks! :pray:

PR #181, opened 2016-03-10
support DOS newlines in NSString.lineAndCharacterForByteOffset(_:)

jpsim:
Fixes realm/SwiftLint#315 /cc @norio-nomura

PR #182, opened 2016-03-13
Use `advancedBy(_:limit:)` on potentially passing endIndex

norio-nomura:
Fix https://github.com/realm/SwiftLint/issues/315
Based on https://github.com/jpsim/SourceKitten/pull/181#issuecomment-195112226


jpsim:
:+1: thanks for doing this, I've been a bit slow these days!

norio-nomura:
Thanks! :pray:

Issue #142, opened 2016-01-21
doc.comment blank for many declarations, causing missing Jazzy docs


pcantrell:
Updating the Siesta docs, I noticed that a bunch of documentation had disappeared. I traced this to Sourcekitten omitting `doc.comment` from many of the declarations. I haven’t verified this, but at a glance, it seems like this might be happening on Swift classes that extend `NSObject`.

You can see the problem in the `Service` and `Resource` classes in Siesta. It shows in the integration specs as they’re currently checked in.

This bug was introduced in jazzy when Sourcekitten was updated to 0.7.3. I didn’t chase it down the rabbit hole any further than that.

pcantrell:
OK, I’ve narrowed it down to a pretty small case, and it’s formatting-dependent. Sourcekitten fails to generate `doc.comment` for this:

    /**
    I am a comment.
    */
    @objc
    public class KittenTest: NSObject { }

…but this works:

    /**
    I am a comment.
    */
    @objc public class KittenTest: NSObject { }

…and so does this:

    /// I am a comment.
    @objc
    public class KittenTest: NSObject { }

jpsim:
Yeah, this is definitely a bug, it's due to the logic used to detect which declarations are documented, it only checks the line directly above (and the logic for `///` might be a bit different than `/** */`).

I wish there was a more straightforward/correct way to get a documentation comment for a declaration using SourceKit. :disappointed:

pcantrell:
FWIW, whatever broke it is a recent addition (0.7.3-ish).

jpsim:
It's not quite as simple as reverting the change... I know what broke this and I added it because it also fixes other things. We need an approach that fixes all of these :smile:

norio-nomura:
It seems we need to fix [here](https://github.com/jpsim/SourceKitten/blob/0.7.4/Source%2FSourceKittenFramework%2FFile.swift#L317).

jpsim:
No, the approach I'm taking there is entirely wrong. I'd rather use the actual Swift parser to detect the doc comment, because it may be many lines above.

norio-nomura:
I understand.

pcantrell:
The “full xml” property in Sourcekitten’s output does work in all cases.

(In some mythical future day when I have more than an hour of free time, I still mean to do a massive, terrifying PR which switches jazzy to using that and rips out its markdown parsing entirely. I say this now just to alarm JP.)

jpsim:
[...]

PR #197, opened 2016-04-07
Fix ignore identifiers starting with '@'

norio-nomura:
`isIdentifier` always returned false.

### DIFF
--- a/Source/SourceKittenFramework/SyntaxMap.swift
+++ b/Source/SourceKittenFramework/SyntaxMap.swift
@@ -69,7 +69,7 @@ public struct SyntaxMap {
             return token.type == SyntaxKind.Identifier.rawValue &&
                 // ignore identifiers starting with '@' because they may be placed between
                 // declarations and their documentation.
-                string?.rangeOfString("@", options: [], range: NSRange(location: token.offset, length: 1)) == nil
+                !(string?.substringWithByteRange(start: token.offset, length: token.length)?.hasPrefix("@") ?? true)
         }
         let isDocOrIdentifier = { isDoc($0) || isIdentifier($0) }
 


jpsim:
Thanks!

norio-nomura:
:pray:

jpsim:
Hey @norio-nomura, do you remember why you did this? In realm/SwiftLint#608 you said that otherwise Swift 2.2 tests failed, but I can't reproduce that, and now we just discovered that this introduced a regression in jazzy: realm/jazzy#530

I obviously don't want to revert this if it will cause other regressions, but I don't know what those are 😄

jpsim:
Never mind, I can reproduce the failures... I'll have to see how to fix one bug without reintroducing the other!

norio-nomura:
Sorry for missing documentation.
That fixes "Stop searching comment on identifier not having `@` prefix." to as is. 
Before fix, that never stop searching on any identifiers.

norio-nomura:
@jpsim I want to see the regression of realm/jazzy#530. Where can I check that?

jpsim:
You can see the diff to jazzy's generated output here: https://github.com/realm/jazzy-integration-specs/compare/jp-render-swift-declarations-in-objc

The regression is the lost extension documentation in Moya.

I have some ideas on how to fix this for both SwiftLint and jazzy but it's a bit complex. It involves checking the structure depth of the previous documentation comment compared to the documented declaration, and validating the structures in between.

norio-nomura:
Thanks for information.
I see following.
```swift
/// Extension for processing raw NSData generated by network access.
@available(*, deprecated, message="This will be removed when ReactiveCocoa 4 becomes final. Please visit https://github.com/Moya/Moya/issues/298 for more information.")
public extension RACSignal {
```
Now I know the reason why you added checking `.Extension` on https://github.com/jpsim/SourceKitten/commit/385115a92957ba2a8b41b4c0ef9437baa394ca72#diff-c40f53a6b61685c6741ff5753303e59aR324.

I think we need to detect attributes much smarter.
I'll have a look about another approach on detecting attribute.

norio-nomura:
> I'll have a look about another approach on detecting attribute.

That seems difficult.
How about separating `getDocumentationCommentBody()` into SwiftLint version and jazzy version?
On jazzy only version, we can use approach of https://github.com/jpsim/SourceKitten/issues/142#issuecomment-173780599

norio-nomura:
I came up with another idea that use parent structure's body offset or sibling structure's end offset for limiting search doc comment.
I'll try to implement that.

jpsim:
> I came up with another idea that use parent structure's body offset or sibling structure's end offset for limiting search doc comment.

Yeah, this is along the lines of what I meant in https://github.com/jpsim/SourceKitten/pull/197#issuecomment-211723098. There's no rush for this, if you don't get around to doing this, I'll probably find time at some point, just to get realm/jazzy#530 out.

norio-nomura:
It seems my fix is not needed.
Thanks!

jpsim:
yeah I fixed this in #95, sorry for not updating this thread when that happened! 😬

PR #198, opened 2016-04-10
[SPM] Specify stricter version requirements of SWXMLHash in Package.swift

norio-nomura:
Because building SWXMLHash-2.3 fails on SPM.
Fixing PR is https://github.com/drmohundro/SWXMLHash/pull/72

SWXMLHash does not have test running SPM on CI, so we should specify stricter version in Package.swift.


jpsim:
:+1: nicely done!

norio-nomura:
Thanks! :pray:

PR #203, opened 2016-05-04
Remove copy framework build phase for SourceKittenFramework target

eyeplum:
The `SourceKittenFramework` target is not depending on `Commandant` and `Result`, and the dependency on `SWXMLHash` is already declared by the `Link Binary with Libraries` build phase. 

We should expect users of the framework adding `SWXMLHash` to their dependency. And then copy both `SourceKittenFramework.framework` and `SWXMLHash.framework` to install path when building their final target, like what we do in the `sourcekitten` target.

This pull request removes the unnecessary copy framework phase in the `SourceKittenFramework` target (which may copy duplicated framework binaries and cause codesigning issues on OS X apps).


eyeplum:
Also, I find that the Swift standard library is copied, too.

Let's say we have an App (called `A.app`) that uses `SourceKittenFramework` (which is also written in Swift), then the result product `A.app` will actually contain two Swift standard libraries.

One located in `A.app/Contents/Frameworks/` and the other one located in `A.app/Contents/Frameworks/SourceKittenFramework.framework/Versions/A/Frameworks/`.

I believe if the "outside one" is identical to the "inside one", the one located inside the framework can actually be deleted to reduce the size of the result App.

eyeplum:
After a little digging, I find that this behavior is controlled by the `EMBEDDED_CONTENT_CONTAINS_SWIFT` build setting. If the target has a embedded content (framework or library) that uses Swift, this should be set to `YES`.

However, it's the same idea as of the library dependencies:
We should expect users of the `SourceKittenFramework` to set `EMBEDDED_CONTENT_CONTAINS_SWIFT = YES`. Otherwise we might make duplicated copy of the Swift standard library.

See: https://developer.apple.com/library/ios/qa/qa1881/_index.html

30218ec fixes this.

norio-nomura:
This PR breaks `make installables`.

jpsim:
Welcome to the joys of shared dynamic frameworks on OS X!

You're right that SourceKittenFramework itself doesn't depend on Commandant and Result, but there's no other parent framework that the `sourcekitten` executable uses. Consider if a user has other tools installed that have `/usr/local/Frameworks` in its runtime search paths, and that also depend on Commandant/Result (e.g. [Carthage](https://github.com/Carthage/Carthage)). Then `carthage` will link our frameworks rather than its own.

Ideally, we would have a separate versioned directory that's not embedded in SourceKittenFramework and that wouldn't conflict with other tools. Like `/usr/local/Frameworks/SourceKitten/...` that would have all our frameworks side by side:

```
SourceKitten/
|-- SourceKittenFramework.framework/
    |-- SWXMLHash.framework/
|-- Commandant.framework/
|-- Result.framework/
```

Like @norio-nomura said, it's critical that whatever changes we make here be compatible with all of SourceKitten's current installation & distribution methods: `.pkg` installer, Homebrew, `Makefile`, Xcode, Swift Package Manager, etc.

eyeplum:
Thanks for the review.

Sorry for not noticing the other installation methods. I was trying to use `SourceKittenFramework` through `Carthage`, I guess a fork is a more suitable choice.

I will close this pull request.

Thanks!

norio-nomura:
I opened #215 that advances this.

PR #224, opened 2016-07-15
[.xcodeproj] Specify project indentation

modocache:
Setting the project's preferred indentation in the project file overrides contributors' defaults, so that no matter their personal preference, they can contribute to SourceKitten with the correct indentation.


jpsim:
Cool stuff. Thanks.

PR #225, opened 2016-07-15
[Index] Support --compilerargs

modocache:
Just as with `sourcekitten complete`, SourceKit provides much better information when given the correct compiler arguments. Provide a `sourcekitten index --compilerargs` argument, to allow for more correct indexing.

---

This is sort of a work-in-progress, because this removes the automatic mechanism for finding the SDK path. Do consumers of SourceKitten rely on this automatic behavior? Should I provide it as a fallback in case the user didn't specify any `--compilerargs`? Advice from maintainers would be appreciated here. :)

### DIFF
--- a/Source/SourceKittenFramework/Request.swift
+++ b/Source/SourceKittenFramework/Request.swift
@@ -191,7 +191,7 @@ public enum Request {
     /// Find USR
     case FindUSR(file: String, usr: String)
     /// Index
-    case Index(file: String)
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks @modocache! Nice work with Nuclide-swift too 😄 . In the future, please add changelog entries according to the format specified in CONTRIBUTING.md.

modocache:
Awesome, will do. Thanks!

PR #227, opened 2016-07-18
[Request] Avoid force casting null responses

modocache:
In cases in which SourceKit returns a null response from an index request (for example, if the index request is not passed any compiler args), a force cast to `[String: SourceKitRepresentable]` would cause a crash at runtime.

Instead, attempt a cast, and if it fails, return an empty dictionary.

### DIFF
--- a/Source/SourceKittenFramework/Request.swift
+++ b/Source/SourceKittenFramework/Request.swift
@@ -328,7 +328,12 @@ public enum Request {
         }
         let response = sourcekitd_send_request_sync(sourcekitObject)
         defer { sourcekitd_response_dispose(response) }
-        return fromSourceKit(sourcekitd_response_get_value(response)) as! [String: SourceKitRepresentable]
+        if let value = fromSourceKit(sourcekitd_response_get_value(response)) {
+          return value as! [String: SourceKitRepresentable]
+        } else {
+          return [String: SourceKitRepresentable]()
+        }
+
     }
     
     /// A enum representation of SOURCEKITD_ERROR_*


Source/SourceKittenFramework/Request.swift
```diff
@@ -328,7 +328,12 @@ public enum Request {
         }
         let response = sourcekitd_send_request_sync(sourcekitObject)
         defer { sourcekitd_response_dispose(response) }
-        return fromSourceKit(sourcekitd_response_get_value(response)) as! [String: SourceKitRepresentable]
+        if let value = fromSourceKit(sourcekitd_response_get_value(response)) {
+          return value as! [String: SourceKitRepresentable]
```
ikesyo: I think this could be just as the following:

```swift
if let value = fromSourceKit(sourcekitd_response_get_value(response)) as? [String: SourceKitRepresentable] {
    return value
} else {
    return [String: SourceKitRepresentable]()
}

```

or `return fromSourceKit(sourcekitd_response_get_value(response)) as? [String: SourceKitRepresentable] ?? [:]`?

norio-nomura:
Sorry for lack of documentation.
We have `failableSend()`.
https://github.com/jpsim/SourceKitten/blob/master/Source/SourceKittenFramework/Request.swift#L375
SwiftLint uses that. 
https://github.com/realm/SwiftLint/blob/62c7d4a992d7783ecb435cd52c556679e00c8edd/Source/SwiftLintFramework/Extensions/File%2BCache.swift#L14

modocache:
@norio-nomura `failableSend()` sounds great, thanks! I encountered this error after modifying the `index` request to take `--compilerargs` as an argument, instead of providing a default -- see #225. With those changes, invoking `sourcekitten index --file /path/to/Source.swift --compilerargs -- badArg` returns a null response, which in turn causes [this `send()`](https://github.com/jpsim/SourceKitten/blob/cbd9625970d968582a218461262a2d70cbb5fb90/Source/sourcekitten/IndexCommand.swift#L36) to crash. Should I change that to a `failableSend()`?

norio-nomura:
`failableSend()` was introduced to survive crashes of SourceKitService such as on linting Swift sources https://github.com/realm/SwiftLint/issues/380.
If we found another failing case of `send()`, we should support that by using `failableSend()` or improved version of that.

modocache:
OK cool, thanks. I'll change #225 to use `failableSend()`, then.

By the way: any thoughts on pull request #225?

jpsim:
#225 looks good to me, I'll take a closer look later today. Thanks for all your recent work on SK lately @modocache! 🙏

modocache:
My pleasure. I really admire the work you and @norio-nomura are doing with SourceKitten!

PR #231, opened 2016-07-29
Change completion's compilerargs to array of strings

keith:
Currently the compilerargs are consumed from commandant as a single
quoted string. Because of this, we have to manually split the string at
spaces, which causes issues with arguments, or filepaths with spaces in
them. Instead of doing this we can allow commandant to pass us the
"remaining" arguments trailing the `--` so we can avoid having to split
anything ourselves.

Currently there is a backwards compatibility problem with this, since now doing:

```
sourcekitten complete ... --compilerargs -- ...
```

Will leave `compilerargs` as an unused argument. If we want to make sure we don't break this for current users we could either try to make this command relevant to the new array of arguments, or we could make it a no-op which at least wouldn't throw an error.

Let me know what you think!

### DIFF
--- a/Source/sourcekitten/CompleteCommand.swift
+++ b/Source/sourcekitten/CompleteCommand.swift
@@ -19,9 +19,9 @@ struct CompleteCommand: CommandType {
         let file: String
         let text: String
         let offset: Int
-        let compilerargs: String
+        let compilerargs: [String]
 
-        static func create(file: String) -> (text: String) -> (offset: Int) -> (compilerargs: String) -> Options {
+        static func create(file: String) -> (text: String) -> (offset: Int) -> (compilerargs: [String]) -> Options {
             return { text in { offset in { compilerargs in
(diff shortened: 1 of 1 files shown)


supermarin:
👍

jpsim:
Looks great, thanks Keith and Marin, and sorry for the delay.

keith:
No worries! Thanks!


--
Keith Smiley

keith:
Should I follow this up with a readme update?


--
Keith Smiley

jpsim:
Readme and changelog ideally pelase.

keith:
Submitted here: https://github.com/jpsim/SourceKitten/pull/241 sorry about the delay!

modocache:
> If we want to make sure we don't break this for current users we could either try to make this command relevant to the new array of arguments, or we could make it a no-op which at least wouldn't throw an error.

This would be awesome, if it's not too much trouble. ❤️

PR #234, opened 2016-08-01
Swift lang syntax to unite types

esuvorov:
Unite swift lang syntax types with SwiftLangSyntax protocol
Make SwiftDocKey public

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@
 
 ##### Enhancements
 
-* None.
+* Refactor to unite swift lang syntax types with SwiftLangSyntax protocol
+* Make SwiftDocKey public
+  [Evgeny Suvorov](https://github.com/esuvorov)
 
 ##### Bug Fixes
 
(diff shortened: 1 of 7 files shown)


jpsim:
@esuvorov thanks for the PR! Can you please explain the rationale behind these changes? I'm cautious when modifying the public API because it affects consumers of the library and restricts what kinds of changes can be made later.

esuvorov:
@jpsim, hi!
I'm currently adding a couple `ASTRule` for [SwiftLint](https://github.com/realm/SwiftLint) which uses _SourceKitten_ as dependency.

As i see _SwiftLint_ framework use a lot of independent strings _"key.everything"_ for AST parsing while in _SourceKitten_ already exist `SwiftDocKey`. Of course i can add similar enum in local module, but as i think much better to use what we have instead of dublicate code.  

Also in _SwiftLint_ exist a couple methods which uses only for `SyntaxKind`, but they could be more usefully and flexible if we change `SyntaxKind` for `SwiftLangSyntax` abstraction.

So i made two main changes:
- `SwiftLangSyntax` protocol for unite _source.lang.swift.*_ family for unification of `StatementKind`, `SwiftDeclarationKind`, `SyntaxKind`. (_I'm not sure about the name of SwiftLangSyntax_)
- and make `SwiftDocKey` public.

jpsim:
Sorry for the long delay @esuvorov. Thanks for the PR, I'll accept this as a short term solution until SourceKitten can move away from untyped dictionaries towards typed declarations (#238).

PR #237, opened 2016-08-20
Support loading `libsourcekitdInProc.so` placed in default search paths

norio-nomura:
Without using `LINUX_SOURCEKIT_LIB_PATH`.


jpsim:
👍 great!

PR #257, opened 2016-09-22
Fix Homebrew bottling error

norio-nomura:
To fix build error on https://github.com/Homebrew/homebrew-core/pull/5094


jpsim:
Thanks for fixing this @norio-nomura! I'll be cutting a new release and updating Homebrew/homebrew-core#5094

PR #254, opened 2016-09-22
Refactor a few things in String+SourceKitten

jpsim:
This may have some performance implications. @norio-nomura what do you think? I pulled this out of #223.

PR #258, opened 2016-09-23
Fix `NSString.lines()` generated surplus line when string ended with new line character

norio-nomura:
This makes it compatible with previous behavior and will fix [failing tests](https://travis-ci.org/realm/SwiftLint/jobs/162080744#L761) of https://github.com/realm/SwiftLint/pull/771.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@
 
 ##### Bug Fixes
 
-* None.
+* `NSString.lines()` generated surplus line when string ended with newline
+  character.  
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the fix! What do you think of #254? Are we better off reverting parts of that instead of this fix?

Otherwise this looks good, thanks for doing this. 👏

norio-nomura:
I'm not sure if it is correct.
Previous behavior could be wrong as I wrote in #259.

jpsim:
Thanks @norio-nomura! 👏

Issue #262, opened 2016-10-09
Error when "make install", impossible to build


itziksi:
After downloading source code,
and running "make install",
I get error, hence impossible to build SourceKitten.

This is the error:

>.../Release/SourceKittenFramework.build/unextended-module.modulemap:2:19: error: umbrella header 'SourceKittenFramework.h' not found
  umbrella header "SourceKittenFramework.h"
                  ^
unknown:0: error: could not build Objective-C module 'SourceKittenFramework'
>** INSTALL FAILED **

norio-nomura:
Current master (e6997b2) requires Xcode 8 for building.
Which version of Xcode did you use?

itziksi:
I've tried both 7.3 and 8.
I now have 7.3 but I have no problem to install and try 8 again.
You don't encounter any problem building a "fresh" clone ?
Thanks for your help !

norio-nomura:
I have no problem building fresh clone of current master (e6997b2) with Xcode 8.
You must have failed with old commit.
Current master (e6997b2) does not support Xcode 7.3.1 because we have migrated SoureKitten to Swift 3.0.
If you want to build Swift 2.3 version on Xcode 7.3.1 or 8, please doing `git checkout 798ce07` before building.

itziksi:
Thanks for your help.
I used "Download" for the source files, so "make install" failed because of missing git repository,
but the error came only in compilation. I think the script that fails is "bootstrap".

Maybe you can add in the README.md, under "Source" : 
_Don’t download the sources as zip, because “make install” requires the git repository_

or break make, if one of the steps fails.

Thanks again.

norio-nomura:
Thanks for investigation. :pray:
I opened #263 for fix this issue by `make bootstrap` will fail when git repository is missing.
```terminal.session
➜  22:11:56  make install
rm -f "SourceKitten.pkg"
rm -rf "/tmp/SourceKitten.dst"
…

script/bootstrap
fatal: Not a git repository (or any of the parent directories): .git
make: *** [bootstrap] Error 128
```

itziksi:
Thanks a lot for your help !

PR #263, opened 2016-10-10
Stop `make` if working tree does not have git repository

norio-nomura:
Fix https://github.com/jpsim/SourceKitten/issues/262
See also: https://github.com/jspahrsummers/objc-build-scripts/issues/13


jpsim:
Thanks!

PR #264, opened 2016-10-12
Update APIs to conform to Swift 3 API Design Guidelines

jpsim:
This _should_ be exhaustive, but it's hard to say without a straightforward way to verify whether or not this satisfies all the suggestions in Apple's design guidelines.

Can you please review @norio-nomura ?


Source/SourceKittenFramework/Request.swift
```diff
@@ -117,83 +116,85 @@ private var sourceKitWaitingRestoredSemaphore = DispatchSemaphore(value: 0)
...
+        return nil
     }
-    return nil
 }
 
 /// Represents a SourceKit request.
 public enum Request {
     /// An `editor.open` request for the given File.
-    case EditorOpen(File)
+    case EditorOpen(file: File)
```
norio-nomura: Do we continue using uppercase member here?

jpsim: you're right, these should probably be lowercased. will do.

norio-nomura:
I have checked public and changed APIs, but not yet on unchanged or unexposed APIs.

jpsim:
Still reviewing this @norio-nomura?

norio-nomura:
Oh, sorry, I got sidetracked to how can we replace `SourceKitRepresentable` with another type.
This is looks good to me. :+1:

jpsim:
Cool, will merge this then and re-target #267 towards `master` so we can move on to that.

PR #267, opened 2016-10-13
Re-add support for Swift Package Manager

jpsim:
Targets and depends on #264.

PR #268, opened 2016-10-13
Support Linux

jpsim:
Targets and depends on #267.

Doesn't build cleanly at the moment due to Commandant pulling in Quick and Nimble when building on Linux, causing conflicts, but the following flow works to build the first time, then `swift build` works afterwards as long as `Packages/` and `.build/` aren't messed with:

```shell
$ swift build
$ rm -rf .build
$ # Delete lines containing 'Quick' in Commandant's Package.swift
$ sed -i.bak '/Quick/d' Packages/Commandant-0.11.1/Package.swift
$ make docker_test
```


jpsim:
The hack I mentioned in my previous comment will no longer be necessary once Carthage/Commandant#80 is included in a release of Commandant.

jpsim:
[Commandant 0.11.2](https://github.com/Carthage/Commandant/releases/tag/0.11.2) is out and I've updated this, so the only thing need to run this is `make docker_test`.

jpsim:
This is ready for review @norio-nomura. Lots of changes in here, a few of which I'm hoping we can improve.

norio-nomura:
Can we add a reference of `Tests/LinuxMain.swift` to Xcode Project with no Target Membership?
Looks good to me other than that. :+1:

Source/SourceKittenFramework/Request.swift
```diff
@@ -57,31 +57,31 @@ private func fromSourceKit(_ sourcekitObject: sourcekitd_variant_t) -> SourceKit
...
         var array = [SourceKitRepresentable]()
-        _ = sourcekitd_variant_array_apply(sourcekitObject) { index, value in
-            if let value = fromSourceKit(value) {
-                array.insert(value, at: Int(index))
-            }
-            return true
+        _ = withUnsafeMutablePointer(to: &array) { arrayPtr in
+            sourcekitd_variant_array_apply_f(sourcekitObject, { index, value, context in
+                if let value = fromSourceKit(value), let context = context {
+                    let localArray = context.bindMemory(to: [SourceKitRepresentable].self, capacity: 1)
```
norio-nomura: Sorry for late notice.
I guess that this can be `assumingMemoryBound(to:)` instead of `bindMemory(to:capacity:)`, because `context` is already initialized before calling `withUnsafeMutablePointer(to:)`.

PR #278, opened 2016-10-22
[experimental] Add UIDNamespace

norio-nomura:
This enables to doing:

``` swift
   if variant.kind == source.lang.swift.decl.function.method.instance,
```

instead of:

``` swift
   if variant.kind == "source.lang.swift.decl.function.method.instance",
```

How do you think about this? @jpsim


norio-nomura:
Performance penalty has been resolved by 2a5f255

jpsim:
Very interesting... would be nice to automate this in codegen like we do for library wrappers

norio-nomura:
> would be nice to automate this in codegen like we do for library wrapper

Already doing. https://github.com/jpsim/SourceKitten/pull/278/files#diff-8ad90e37080bd05c47b0e531e05a4251R16

norio-nomura:
Oh, sorry, I misunderstood! never mind last comment. 🙇

Tests/SourceKittenFrameworkTests/VariantPerformanceTests.swift
```diff
@@ -128,6 +128,26 @@ class VariantPerformanceTests: XCTestCase {
         }
         XCTAssertEqual(availables.sorted(), self.expectedAvailables)
     }
+
+    func testFindAvailablesWithVariantUIDNamespace() {
+        func findAvailables(variant: SourceKitVariant) -> [String] {
+            let resultFromSubstructure = variant.subStructure?.flatMap(findAvailables) ?? []
+            if variant.kind == UID.source.lang.swift.decl.function.method.instance,
+                let attributes = variant.attributes?.flatMap({ $0.attribute }),
+                attributes.contains(.available),
```
norio-nomura: Now, some properties of `SourceKitVariant` returns specific types of `UIDNapespace`.
e.g. `attribute` returns `UID.source.decl.attribute`, so it can be equatable with `.available` instead of `UID.source.decl.attribute.available`.

norio-nomura:
I'll write matching pattern tests of `UIDNamespace` types.

norio-nomura:
I changed nested namespaces to flat namespaces.
Because nested namespaces can’t support coexistence both of `"source.lang.swift.syntaxtype.comment"` and `"source.lang.swift.syntaxtype.comment.mark"`.

jpsim:
@norio-nomura, hmm I'm still not convince about this approach. It shows promise, but I haven't thoroughly reviewed it. We usually wait for each other to review large PRs, so I'm curious why you merged this one? Maybe it was an accident? Or maybe I'm missing something.

jpsim:
Oops, nevermind, I was missing something, this was merged to `nn-sourcekit-variant`. Please ignore my last comment 😅

norio-nomura:
Sorry for surprising you 🙇 I should wrote some comment.

jpsim:
Not at all. Entirely my mistake.

Issue #243, opened 2016-09-15
Objective-C class properties cause SIGILL


frankus:
Declaring a class property in Objective-C:

```
@property (class, readonly, nonatomic) Foo *shared;
```

and running sourcekitten (arguments of `doc -objc -- -x objective-c -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator10.0.sdk -I /path/to/project`) against it fails with a `terminated by signal SIGILL (Illegal instruction)` error.

jpsim:
huh, interesting... thanks for reporting this

jeremiegirault:
more info here: 
`private func accessorUSR(getter: Bool) -> String` checks for `py`
class properties contains a `cpy` tag
force unwrap causes the crash

jeremiegirault:
more: proposed a fix in #285

jpsim:
Merged in #291. Thanks to @jeremiegirault and @frankus for reporting 😄

PR #285, opened 2016-10-27
fix class properties by parsing the good tag

jeremiegirault:
handles #243

I don't know how sourcekitten works so this must be a dirty patch, but I had to fix it quickly to make jazzy work for me. 

Better solutions are welcome but it's working (not crashing at least) and is a good starting point.

### DIFF
--- a/Source/SourceKittenFramework/SourceDeclaration.swift
+++ b/Source/SourceKittenFramework/SourceDeclaration.swift
@@ -71,7 +71,7 @@ public struct SourceDeclaration {
         guard let declaration = declaration else {
             fatalError("Couldn't extract declaration")
         }
-        let pyStartIndex = usr.range(of: "(py)")!.lowerBound
+        let pyStartIndex = usr.range(of: "(py)")?.lowerBound ?? usr.range(of: "(cpy)")!.lowerBound
         let usrPrefix = usr.substring(to: pyStartIndex)
         let fullDeclarationRange = NSRange(location: 0, length: (declaration as NSString).length)
         let regex = try! NSRegularExpression(pattern: getter ? "getter\\s*=\\s*(\\w+)" : "setter\\s*=\\s*(\\w+:)", options: [])
@@ -80,7 +80,7 @@ public struct SourceDeclaration {
             let accessorName = (declaration as NSString).substring(with: matches[0].rangeAt(1))
             return usrPrefix + "(im)\(accessorName)"
         } else if getter {
-            return usr.replacingOccurrences(of: "(py)", with: "(im)")
+            return usr.replacingOccurrences(of: "(py)", with: "(im)").replacingOccurrences(of: "(cpy)", with: "(im)")
         }
         // Setter
         let capitalFirstLetter = String(usr.characters[usr.characters.index(pyStartIndex, offsetBy: 4)]).capitalized
--- a/Source/SourceKittenFramework/SourceDeclaration.swift
+++ b/Source/SourceKittenFramework/SourceDeclaration.swift
@@ -71,21 +71,49 @@ public struct SourceDeclaration {
         guard let declaration = declaration else {
             fatalError("Couldn't extract declaration")
         }
-        let pyStartIndex = usr.range(of: "(py)")?.lowerBound ?? usr.range(of: "(cpy)")!.lowerBound
+        enum AccessorType {
+            case `class`, instance
+
+            var propertyTypeString: String {
+                switch self {
+                case .class: return "(cpy)"
+                case .instance: return "(py)"
+                }
+            }
+
+            var methodTypeString: String {
+                switch self {
+                case .class: return "(cm)"
+                case .instance: return "(im)"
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for investigating this and finding a fix! This will need a bit more work before it can land, though.

First, I'd like to see a unit test that covers this change. For example, the existing `ClangTranslationUnitTests.testBasicObjectiveCDocs()` test could be extended to cover this by adding the following to the `Musician.h` fixture:

```diff
 #pragma mark - Properties
 
+/// Does this class model musicians?
+@property (class, readwrite, nonatomic) BOOL isMusician;
+
 /**
  The name of the musician. i.e. "John Coltrane"
  */
```

By applying this change, I noticed that clang generates 3 declarations when declaring a readwrite property (or 2 if readonly):

```json
{
  "key.filepath" : "Musician.h",
  "key.doc.file" : "Musician.h",
  "key.deprecation_message" : "",
  "key.always_unavailable" : false,
  "key.unavailable_message" : "",
  "key.parsed_scope.start" : 17,
  "key.kind" : "sourcekitten.source.lang.objc.decl.property",
  "key.swift_declaration" : "class var isMusician: Bool { get set }",
  "key.doc.full_as_xml" : "",
  "key.always_deprecated" : false,
  "key.doc.line" : 17,
  "key.doc.column" : 46,
  "key.name" : "isMusician",
  "key.doc.comment" : "Always returns `YES`.",
  "key.usr" : "c:objc(cs)JAZMusician(cpy)isMusician",
  "key.parsed_declaration" : "@property (assign, readwrite, nonatomic, class) BOOL isMusician;",
  "key.parsed_scope.end" : 17
},
{
  "key.filepath" : "Musician.h",
  "key.doc.file" : "Musician.h",
  "key.deprecation_message" : "",
  "key.always_unavailable" : false,
  "key.unavailable_message" : "",
  "key.parsed_scope.start" : 17,
  "key.kind" : "sourcekitten.source.lang.objc.decl.method.class",
  "key.always_deprecated" : false,
  "key.doc.line" : 17,
  "key.doc.column" : 46,
  "key.name" : "+isMusician",
  "key.usr" : "c:objc(cs)JAZMusician(cm)isMusician",
  "key.parsed_declaration" : "+ (BOOL)isMusician;",
  "key.parsed_scope.end" : 17
},
{
  "key.filepath" : "Musician.h",
  "key.doc.file" : "Musician.h",
  "key.deprecation_message" : "",
  "key.always_unavailable" : false,
  "key.unavailable_message" : "",
  "key.parsed_scope.start" : 17,
  "key.kind" : "sourcekitten.source.lang.objc.decl.method.class",
  "key.always_deprecated" : false,
  "key.doc.line" : 17,
  "key.doc.column" : 46,
  "key.name" : "+setIsMusician:",
  "key.usr" : "c:objc(cs)JAZMusician(cm)setIsMusician:",
  "key.parsed_declaration" : "+ (void)setIsMusician:(BOOL)isMusician;",
  "key.parsed_scope.end" : 17
}
```

So that's the property, along with getter and setter methods. For instance properties, we reject implicitly generated property getters & setters (see [Sequence<SourceDeclaration>.rejectPropertyMethods()](https://github.com/jpsim/SourceKitten/blob/6b7a0fe2f4572de5464093a484044707e9438b00/Source/SourceKittenFramework/SourceDeclaration.swift#L113-L123)).

However, because `accessorUSR(getter:)` returns incorrect results for class properties, this doesn't work for those. It generates incorrect results for a few reasons:

* You'll notice in the JSON snippet I've shared above, that the USR uses `(cm)` to indicate a "class method" rather than the `(im)` you're using, which denotes an "instance method". The fix here is to change `replacingOccurrences(of: "(cpy)", with: "(im)")` to `replacingOccurrences(of: "(cpy)", with: "(cm)")`
* `restOfSetterName` is incorrect because `restOfSetterName` assumes a length of `(py)` (4), which is incorrect for `(cpy)` (5).

I've pushed up my partial work to address these issues in 6499d8b, but that could certainly be improved by some refactoring. @jeremiegirault if you're up for it, I'll let you continue your work in this PR to address the points I've brought up.

Thanks again for the contribution.

jeremiegirault:
I'll look at this.

jpsim:
Let me know if you have any questions or would like any help.

jeremiegirault:
sure !

jeremiegirault:
your fix looks good, just rebased your master and updated some fixtures to have tests passing, is it good ?

jpsim:
Looks like you made some changes to account for Xcode 8.1 and Swift 3.0.1, but since the unit tests still run on 8.0, they're now failing.

jeremiegirault:
ok then I'll revert this

jeremiegirault:
done

jpsim:
Thanks for your hard work on this @jeremiegirault! Will merge in #291.

jeremiegirault:
haha it was mostly your work @jpsim but thank you :)

PR #287, opened 2016-10-28
Add unit test for SPM docs

jpsim:
YamlSwift takes over a minute to parse `debug.yaml` on Linux! As soon as I figure out how to profile Swift on Linux for performance, I'll look into this more.


norio-nomura:
Swift Package Manager always uses debug configuration. We need to check performance with release configuration.

jpsim:
> Swift Package Manager always uses debug configuration. We need to check performance with release configuration.

Same 😞

PR #293, opened 2016-11-06
remove 'pod repo update' from .travis.yml

jpsim:
shouldn't be needed anymore since versions are specified in the podspec


jpsim:
nope, still needed it appears

PR #294, opened 2016-11-07
some fixes for xcode 8.1

jeremiegirault:
I noticed that I made a typo in a code comment of mine and typed `@type` instead of `@param`.
This was interpreted as a comment command. Previously `clang_BlockCommandComment_getCommandName` returned something. With Xcode 8.1 I had a force unwrap.
I noticed the value was actually returned by `clang_InlineCommandComment_getCommandName`

Maybe we should keep the force unwrap instead of the default value (`""`), but do it in `commandName` if it is always expected to return a value ? It could allow us to detect issues like these.

I also updated the fixtures for Xcode 8.1

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,11 @@
 
 ##### Bug Fixes
 
-* `CXComment.commandName()` was returning nil on custom code comments 
-  since xcode 8.1 sdk. This caused a force unwrap when generating 
-  documentation. Inline command comment is now used as a 
-  fallback to catch this edge case.
+* `CXComment.commandName()` was returning nil on custom code comments
+  since Xcode 8.1. This caused a force unwrap when generating
+  documentation. Inline command comment is now used as a
+  fallback to catch this edge case.  
+  [Jérémie Girault](https://github.com/jeremiegirault)
 
 ## 0.15.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@
 
 ##### Bug Fixes
 
-* None.
+* `CXComment.commandName()` was returning nil on custom code comments 
+  since xcode 8.1 sdk. This caused a force unwrap when generating 
+  documentation. Inline command comment is now used as a 
+  fallback to catch this edge case.
 
 ## 0.15.0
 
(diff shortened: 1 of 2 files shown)


Source/SourceKittenFramework/Clang+SourceKitten.swift
```diff
@@ -255,7 +256,8 @@ extension CXComment {
     }
 
     func commandName() -> String? {
-        return clang_BlockCommandComment_getCommandName(self).str()
+        return clang_InlineCommandComment_getCommandName(self).str() ??
```
jpsim: shouldn't we return 'block' instead of 'inline' to preserve the previous behavior?

Source/SourceKittenFramework/library_wrapper_sourcekitd.swift
```diff
@@ -9,7 +9,7 @@ private let path = "sourcekitd.framework/Versions/A/sourcekitd"
 private let library = toolchainLoader.load(path: path)
 internal let sourcekitd_initialize: @convention(c) () -> () = library.load(symbol: "sourcekitd_initialize")
 internal let sourcekitd_shutdown: @convention(c) () -> () = library.load(symbol: "sourcekitd_shutdown")
-internal let sourcekitd_set_interrupted_connection_handler: @convention(c) (sourcekitd_interrupted_connection_handler_t) -> () = library.load(symbol: "sourcekitd_set_interrupted_connection_handler")
+internal let sourcekitd_set_interrupted_connection_handler: @convention(c) (@escaping sourcekitd_interrupted_connection_handler_t) -> () = library.load(symbol: "sourcekitd_set_interrupted_connection_handler")
```
jpsim: we shouldn't update these fixtures for Xcode 8.1 until we can run that version on Travis. (and Swift 3.0.1 on Linux).

jeremiegirault:
Right, I reverted the Xcode 10.1 features and reordered the order of calls in `commandName` in order to preserve original behaviour.

jpsim:
Could you please add a changelog entry? Thanks!

jeremiegirault:
Hey, thanks for the feedback ! I added a small entry in the changelog.

CHANGELOG.md
```diff
@@ -40,6 +40,8 @@
 
 ##### Bug Fixes
 
+* `CXComment.commandName()` was returning nil on custom code comments since xcode 8.1 sdk. This caused a force unwrap when generating documentation. Inline command comment is now used as a fallback to catch this edge case.
```
jpsim: This is in the wrong changelog section. An empty changelog section was added in 420b07a which should contain this entry. Try rebasing this PR to be able to move this entry to that section.

jpsim: Also, we hard-wrap CHANGELOG lines at 80 characters. See [CONTRIBUTING.md](https://github.com/jpsim/SourceKitten/blob/master/CONTRIBUTING.md).

jeremiegirault:
oops sorry. Edited

jpsim:
Thanks @jeremiegirault 🙏

Issue #299, opened 2016-11-16
Set APPLICATION_EXTENSION_API_ONLY for SourceKittenFramework and all its dependencies


jpsim:
Which is necessary to use SourceKitten in an Xcode Source Editor Extension

norio-nomura:
Application Extension requires sandboxing, AFAIK. Did you find any workaround other than using xpc?

klaaspieter:
Is it possible to accidentally use the framework in a way that wouldn't pass review? Should that be documented somewhere?

PR #300, opened 2016-11-17
Allow use in app extensions

klaaspieter:
Waiting for behrang/YamlSwift#45 and drmohundro/SWXMLHash#106 to be merged and released.

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,5 +1,5 @@
 github "antitypical/Result" "3.0.0"
-github "drmohundro/SWXMLHash" "3.0.2"
-github "behrang/YamlSwift" "3.1.0"
+github "drmohundro/SWXMLHash" "3.0.3"
+github "behrang/YamlSwift" "3.2.0"
 github "jspahrsummers/xcconfigs" "3576b438c47dc1a118672ab81c6d916322b062e8"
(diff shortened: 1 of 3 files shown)


klaaspieter:
Instead of waiting for the frameworks to be released we could point Carthage at the branches on my forks temporarily.

I'm also [still wondering][] if this allows users to use non-extension-safe APIs.

[still wondering]: https://github.com/jpsim/SourceKitten/issues/299#issuecomment-261251996

jpsim:
> I'm also still wondering if this allows users to use non-extension-safe APIs.

It's not that the APIs aren't extension safe, it's that SourceKitten opens an XPC connection to `sourcekitd.framework` with any codepath that goes through `Request.send()` basically.

klaaspieter:
Can we prevent that somehow? Should it be documented somewhere?

klaaspieter:
The last two commits should be everything that was left to do for this to work.

jpsim:
> Can we prevent that somehow? Should it be documented somewhere?

Don't know, and probably. This would need a bit more research. Maybe @norio-nomura has some ideas here since he's actually used SourceKitten in an Xcode Source Editor Extension already. I haven't, so I'm not super familiar.

jpsim:
Looks great, thanks!

PR #307, opened 2016-12-03
Reduce build time

norio-nomura:
- Add `display_compilation_time` target to `Makefile`
- Reduce build time of two functions
- Set `-Xfrontend -warn-long-function-bodies=200` to `OTHER_SWIFT_FLAGS`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,3 +74,7 @@ release: package archive
 
 docker_test:
 	docker run -v `pwd`:/SourceKitten norionomura/sourcekit:301 bash -c "cd /SourceKitten && swift test"
+
+# http://irace.me/swift-profiling/
+display_compilation_time:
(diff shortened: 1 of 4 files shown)


jpsim:
Thanks! We should probably lower the threshold even more if we can things to be stable (100-150).

norio-nomura:
I think [The failing test on SPM](https://travis-ci.org/jpsim/SourceKitten/jobs/180907909) is affected by version up of `Result` that updated from 3.0.0 to 3.1.0.
I'll push a PR fixing that.

norio-nomura:
Oh, [test triggered by merge commit](https://travis-ci.org/jpsim/SourceKitten/jobs/180910543) succeeded…

norio-nomura:
> Oh, test triggered by merge commit succeeded…

Sorry, I started this PR from old master. 🙇

PR #310, opened 2016-12-08
Set 'use legacy swift language version' to 'no' (== v3.0.1)

Johennes:
This is needed in order to be able to link against the framework
when building it through Carthage on Xcode 8.1 (8B62).


jpsim:
Thanks!

Issue #312, opened 2016-12-09
Support arbitrary sourcekit command?


keith:
While working on some editor integrations I found myself wanting to use sourcekit to get some info about my files/modules. To do this I was planning on passing some compiler arguments, along with some other info, to sourcekit for various requests that aren't particularly supported with a custom command in sourcekitten.

What do you all think about adding the ability in sourcekitten to execute arbitrary sourcekit requests, rather than the ones that are already supported with custom arguments in sourcekitten, so that you could pass something like:

```
{
  key.request: source.request.cursorinfo,
  key.compilerargs: [
      ...
  ],
  key.offset: 123,
  key.sourcefile: "path/to/foo.swift"
}
```

through to sourcekit, without having to add a custom command in sourcekitten to handle this directly for every possible current, and future, sourcekit request. The benefit being you'd still be able to utilize the communication and parsing handled in sourcekitten, without having to have a command for each possible request.

keith:
Another possible solution to this problem would be for me to just use `Request.customRequest` from my own CLI/tool, and not from sourcekitten's CLI

keith:
Even better I could use `Request.customRequest` along with `sourcekitd_request_create_from_yaml` to do this pretty simply with some yaml files. But I'm still curious if you think this has a place in sourcekitten itself.

jpsim:
Great idea! As you mentioned, the `Request.customRequest` is the right API for this, although a CLI command would be great.

keith:
I just submitted a slightly related PR which may or may not make sense to merge: https://github.com/jpsim/SourceKitten/pull/313 but this would allow consumers (possible for this usage) to pass yaml as well.

jpsim:
nope, I think that's a great way to expose it

keith:
Awesome! I think this is definitely a good start since this would also mean we wouldn't have to convert the jsonish representation, instead we could just pass the yaml right through

PR #313, opened 2016-12-09
Add ability to create request from yaml

keith:
This allows consumers of SourceKittenFramework to pass a yaml formatted
string in order to create a sourcekit request.

Related: https://github.com/jpsim/SourceKitten/issues/312


jpsim:
A tiny test for this would be 👌 💯

keith:
Can do!

keith:
Updated with a test. If you think this test should move somewhere else let me know, I wrestled with where to put it.

jpsim:
> I wrestled with where to put it.

`SourceKitTests.swift` is the best place for it 👍. Add a changelog entry and :shipit:

keith:
Added entry!

jpsim:
Thanks @keith!

jpsim:
Next up, you should expose this in the CLI 😉

keith:
That is the plan! What do you think about either:

```
sourcekitten yaml --file foo.yml (or --text "yaml string")
```

Or maybe `yaml-request` instead of just `yaml`?

jpsim:
I like `yaml-request`. Or `sourcekitten request --yaml [file or yaml string]`

keith:
Submitted here: https://github.com/jpsim/SourceKitten/pull/315

PR #316, opened 2016-12-13
Add 'SWIFT_VERSION=3.0' to project file

S2dentik:
It doesn't build with Xcode 8.2 otherwise.

### DIFF
--- a/sourcekitten.xcodeproj/project.pbxproj
+++ b/sourcekitten.xcodeproj/project.pbxproj
@@ -784,6 +784,7 @@
 				LD_RUNPATH_SEARCH_PATHS = "@loader_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = "com.sourcekitten.$(PRODUCT_NAME:rfc1034identifier)";
 				PRODUCT_NAME = SourceKittenFramework;
+				SWIFT_VERSION = 3.0;
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
 				WARNING_CFLAGS = (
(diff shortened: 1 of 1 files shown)


norio-nomura:
Should we change them on project configuration instead of target configurations?

jpsim:
☝️ what @norio-nomura said.

Thanks @S2dentik!

S2dentik:
Yeah, that would make sense, tried that, but Xcode doesn't actually take into consideration those, nor are we able to specify `$(inherited)` for `SWIFT_VERSION` so adding it to project configuration doesn't fix the thing.

S2dentik:
Another question would be, should we also change the `travis.yml`?

norio-nomura:
Current project config already has `SWIFT_VERSION = 3.0.1`. We don't need `$(inherited)` on them.
> Another question would be, should we also change the `travis.yml`?

I have no objection to that.

jpsim:
Closing since there's no need to do this given that the current project config already has `SWIFT_VERSION = 3.0.1` as @norio-nomura said. I'll change travis to test on Xcode 8.2 in an upcoming PR.

S2dentik:
@jpsim This is an Xcode bug, and that's what this PR actually does, it fixes that.
http://stackoverflow.com/questions/41118912/use-legacy-swift-language-version-xcode-8-2
In the new version of Xcode you should basically select the `Use Legacy Swift version` to `No` again, and it does add this line to target. Without it, the target doesn't build.

jpsim:
@S2dentik yeah, sorry I misunderstood, but fixing this at the project level is still better than at the framework level. See #317.

S2dentik:
@jpsim Yeah, I guess changing it on the project level would be a better decision 👍 Totally agree

PR #325, opened 2016-12-27
Rename Releasing.md to RELEASING.md

istx25:
`Releasing.md` looked out of place when `LICENSE` and `README.md` were both uppercased. So I thought I'd make it consistent! 💎 🎈


jpsim:
I appreciate the thought @istx25, but this is the convention I've followed with several projects, and have personally always disliked the all-caps conventions for changelog and readme (seems overly aggressive), but conformed to it because the vast majority of projects do. So for new file names that we introduce, I'd prefer sticking to something less "screamy" 😄 .

istx25:
hahaha okay no worries. that's a fair objection. 

IT DOES SOUND A LITTLE SCREAMY.

:)

PR #327, opened 2017-01-06
Add locks around shared global state variables

krzysztofzablocki:
Global state dictionaries aren't implicitly thread safe so added locks around access to those variables, so that we avoid crashes in swiftlint and sourcery when using multi-threading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@
 
 ##### Bug Fixes
 
-* Added locks around global shared state variables to avoid multi-threading problems.[@krzysztofzablocki](https://github.com/krzysztofzablocki)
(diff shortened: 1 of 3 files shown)


jpsim:
This is a good step in the right direction! Thanks.

But note that this doesn't make SourceKitten fully thread-safe yet, I think. For example, I'm pretty sure the way we [cache some String info with ObjC associated objects](https://github.com/jpsim/SourceKitten/blob/0.16.0/Source/SourceKittenFramework/String+SourceKitten.swift#L194-L205) isn't thread safe.

krzysztofzablocki:
I wonder if that one will affect Sourcery because we parallelize at File level, so I just need to have thread safety across instances of SourceKitten structures, not at object level

PR #328, opened 2017-01-07
Use ptthread_mutex instead of locks for significant performance boost

krzysztofzablocki:
As I'm doing more work on threading I have seen that `NSLock` is really slow, using pt_thread_mutex lowers my performance metric from 3.5s to 2.6. No functionality change here.

PS. SpinLock might be faster but it has many side-effects, so I preferred to use safer yet efficent approach here.

### DIFF
--- a/Source/SourceKittenFramework/Clang+SourceKitten.swift
+++ b/Source/SourceKittenFramework/Clang+SourceKitten.swift
@@ -15,19 +15,19 @@ import Foundation
 import SWXMLHash
 
 private var _interfaceUUIDMap = [String: String]()
-private var _interfaceUUIDMapLock = NSLock()
+private var _interfaceUUIDMapMutex = pthread_mutex_t()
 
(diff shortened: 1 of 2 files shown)


jpsim:
Looking at the corelibs implementation, this is exactly what `NSLock` uses under the hood: https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSLock.swift

And if the Darwin Foudnation version is slower, there's probably a good reason for it.

krzysztofzablocki:
Not really, there is overhead due to closures and indirection(that's why I didn't refactor repetition in the original locking), ptthread isn't magic it's industry standard, you can read more about this at Matt blog https://www.cocoawithlove.com/blog/2016/06/02/threads-and-mutexes.html#conclusion

krzysztofzablocki:
Ok I made a mistake here, the difference is much smaller, closing this :)

PR #329, opened 2017-01-07
Add option to disable uid cache for requests

krzysztofzablocki:
Adds option to disable request global UID cache, this might be useful if someone wants to use SourceKittenFramework to parse specific files in parallel, e.g. Sourcery.


jpsim:
Lets just disable always: see #330. Thanks for prompting me to look into this though!

PR #332, opened 2017-02-03
update Yams

jpsim:


norio-nomura:
My patch against `Module.swift` is following:
```diff
diff --git a/Source/SourceKittenFramework/Module.swift b/Source/SourceKittenFramework/Module.swift
index a8117a1..ebe4d4d 100644
--- a/Source/SourceKittenFramework/Module.swift
+++ b/Source/SourceKittenFramework/Module.swift
@@ -36,47 +36,15 @@ public struct Module {
 
     public init?(spmName: String) {
         let yamlPath = ".build/debug.yaml"
-        guard let yaml = try? Yams.Node(string: String(contentsOfFile: yamlPath, encoding: .utf8)),
-            case let .mapping(yamlMapping) = yaml
+        guard let yaml = try? Yams.compose(yaml: String(contentsOfFile: yamlPath, encoding: .utf8))
             else { fatalError("SPM build manifest does not exist at `.build/debug.yaml` or does not match expected format.") }
 
-        guard let yamlCommands = yamlMapping.first(where: { $0.0 == "commands" })?.1,
-            case let .mapping(yamlCommandsMapping) = yamlCommands
+        guard let moduleCommand = yaml?["commands"]?["<SourceKittenFramework.module>"],
+            let imports = moduleCommand["import-paths"]?.array(of: String.self),
+            let otherArguments = moduleCommand["other-args"]?.array(of: String.self),
+            let sources = moduleCommand["sources"]?.array(of: String.self)
             else { fatalError("SPM build manifest does not match expected format.") }
 
-        guard let moduleCommands = yamlCommandsMapping.first(where: { $0.0 == "<\(spmName).module>" })?.1,
-            case let .mapping(moduleCommand) = moduleCommands
-            else {
-                fputs("Could not find SPM module '\(spmName)'. Here are the modules available:\n", stderr)
-                let availableModules: [String] = yamlCommandsMapping.flatMap { command in
-                    if case let .mapping(commandMapping) = command.1,
-                        let optionalCommandName = commandMapping.first(where: { $0.0 == "module-name" })?.1,
-                       case let .scalar(commandName) = optionalCommandName {
-                        return commandName
-                    }
-                    return nil
-                }
-                fputs("\(availableModules.map({ "  - " + $0 }).joined(separator: "\n"))\n", stderr)
-                return nil
-        }
-
-        func stringArray(_ key: String) -> [String]? {
-            if let seq = moduleCommand.first(where: { $0.0 == key })?.1,
-                case let .sequence(array) = seq {
-                return array.flatMap { scalar in
-                    if case let .scalar(string) = scalar {
-                        return string
-                    }
-                    return nil
-                }
-            }
-            return nil
-        }
-        guard let imports = stringArray("import-paths"),
-              let otherArguments = stringArray("other-args"),
-              let sources = stringArray("sources") else {
-                return nil
-        }
         name = spmName
         compilerArguments = {
             var arguments = sources
```

norio-nomura:
My patch uses `Yams.compose()` that uses `Node` instead of `[String:Any]`, and is faster than `Yams.load()`.

norio-nomura:
Sorry, I found bug in my patch.
I'll update them.

Source/SourceKittenFramework/Module.swift
```diff
@@ -36,46 +36,20 @@ public struct Module {
...
-            if let seq = moduleCommand.first(where: { $0.0 == key })?.1,
-                case let .sequence(array) = seq {
-                return array.flatMap { scalar in
-                    if case let .scalar(string) = scalar {
-                        return string
-                    }
-                    return nil
-                }
-            }
+        guard let moduleCommand = yamlCommands.first(where: { ($0["module-name"] as? String ?? "") == spmName }) else {
```
norio-nomura: Are there any reason to change from checking  "key is `"<\(spmName).module>"`"?

jpsim: We used to check for `module-name` with YamlSwift, but moving to the previous crappy Yams interface it was hard to do, so I switched it to this, but I suspect there are situations in which the module name is modified when it's in the `<*.module>` key.

PR #334, opened 2017-02-06
Added numBytesToErase to CodeCompletionItem

KelvinJin:
This field is defined in the [protocol](https://github.com/apple/swift/blob/master/tools/SourceKit/docs/Protocol.md) and will be pretty useful to have in order to do autocompletion in an editor.

Tests passed locally on my mac. If it looks good, I'll update the change log too.


KelvinJin:
hmm, seems tests failed under linux. Not sure why though.

Source/SourceKittenFramework/CodeCompletionItem.swift
```diff
@@ -26,17 +26,19 @@ public struct CodeCompletionItem: CustomStringConvertible {
     public let moduleName: String?
     public let docBrief: String?
     public let associatedUSRs: String?
+    public let numBytesToErase: Int64?
```
norio-nomura: Changing this to `Int` will fix crash in `CodeCompletionTests` on Linux.

KelvinJin: Hmm, that's annoying. Have to add a typealias. I'm bad at naming.

KelvinJin:
All right. Should be cool now.

jpsim:
Thanks!

PR #335, opened 2017-02-07
Change packaging

norio-nomura:
- Stop using `swift-stdlib-tool`  
  Xcode 8.3 beta 2 no longer includes `swift-stdlib-tool`.
  So, we need changes:
  - Remove “Embed Swift libraries into SourceKittenFramework.framework” build phase that using `swift-stdlib-tool`
  - Add a job to `installables` target that moves Swift Standard Libraries from `sourcekitten.app` bundle to `SourceKittenFramework.framework`
- Change so that `SourceKitten.pkg` does not include `/Applications`

The former change is also needed in `SwiftLint` on Xcode 8.3 beta 2.


norio-nomura:
Thanks! 🙏

PR #341, opened 2017-03-01
Expose properties of `SourceDeclaration` which can be used by `SourceKittenFramework` users.

ainopara:
I'm making a command line tool to generate documentation from Objective-C headers. (Similar to jazzy but written in swift). So I embed `SourceKittenFramework` in my app and use `ClangTranslationUnit` to parse headers and get `SourceDeclaration` instances. If those properties can not be access outside the framework, it seems I must transform `SourceDeclaration` instances to JSON then parse JSON to Dictionary then map to my models.

By mark those properties public, application can directly map information from `SourceDeclaration` to their models. Rather than parse the JSON output from `ClangTranslationUnit`.

### DIFF
--- a/Source/SourceKittenFramework/Clang+SourceKitten.swift
+++ b/Source/SourceKittenFramework/Clang+SourceKitten.swift
@@ -39,11 +39,11 @@ struct ClangIndex {
     }
 }
 
-struct ClangAvailability {
-    let alwaysDeprecated: Bool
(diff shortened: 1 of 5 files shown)


jpsim:
Looks good, @ainopara! Is this doc generation tool being built in the open? I'd love to see it!

ainopara:
I will make it public when it's ready. 😉

PR #344, opened 2017-03-16
Avoid force unwrapping crash

toshi0383:
I encountered crash when using Sourcery.
It worked fine after applying this fix.
The content of line was a Japanese source code comment, and `samePosition(in:)` was returning nil.


toshi0383:
Somehow I cannot reproduce the crash any more.😕
Closing this for now.

PR #349, opened 2017-03-28
force cast of optional

0xPr0xy:
swift 3.1


norio-nomura:
Thanks for contribution!
As far as I tested on #352 , this fix was unnecessary.

0xPr0xy:
Closing this, tested with the swift3.1 branch and it does not error out here.
It does throw this error on xcodebuild test though

```
SourceKittenFramework/Module.swift:10:8: no such module 'Yams'
```

PR #351, opened 2017-03-28
Fix the trimming of trailing characters.

### LINKED ISSUE #350: SourceKitten crashes when a documentation comment ends with an extended grapheme cluster.
NSString and Swift's String handle extended grapheme clusters differently. While removing trailing whitespace, SourceKitten tries to create a UnicodeScalar from an NSString character which does not necessarily succeed. Because of that SourceKitten crashes at force-unwrapping the unicode scalar (`String+SourceKitten.swift`, line 241 in `trimmingTrailingCharacters(in characterSet: CharacterSet)`.

Minimal example:
```Swift
/**💣*/
class Foobar{
}
```

Lukas-Stuehrk:
When the string ended with extended grapheme clusters, the method
crashed because of the force-unwrapping. This refactoring adds support
for handling extended grapheme clusters.

This fixes #350

### DIFF
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -232,14 +232,15 @@ extension NSString {
     - parameter characterSet: Character set to check for membership.
     */
     public func trimmingTrailingCharacters(in characterSet: CharacterSet) -> String {
-        if length == 0 {
+        guard length > 0 else {
             return ""
         }
-        var charBuffer = [unichar](repeating: 0, count: length)
-        getCharacters(&charBuffer, range: NSRange(location: 0, length: charBuffer.count))
-        for newLength in (1...length).reversed() {
-            if !characterSet.contains(UnicodeScalar(charBuffer[newLength - 1])!) {
-                return substring(with: NSRange(location: 0, length: newLength))
+        var string = self.bridge()
+        let endIndex = string.unicodeScalars.endIndex
+        for offset in 1...string.unicodeScalars.count {
+            let index = string.unicodeScalars.index(endIndex, offsetBy: -offset)
+            if !characterSet.contains(string.unicodeScalars[index]), let stringIndex = index.samePosition(in: string) {
+                return string.substring(to: string.index(stringIndex, offsetBy: 1))
             }
         }
         return ""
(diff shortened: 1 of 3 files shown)


Source/SourceKittenFramework/String+SourceKitten.swift
```diff
@@ -232,14 +232,15 @@ extension NSString {
...
         }
-        var charBuffer = [unichar](repeating: 0, count: length)
-        getCharacters(&charBuffer, range: NSRange(location: 0, length: charBuffer.count))
-        for newLength in (1...length).reversed() {
-            if !characterSet.contains(UnicodeScalar(charBuffer[newLength - 1])!) {
-                return substring(with: NSRange(location: 0, length: newLength))
+        var string = self.bridge()
+        let endIndex = string.unicodeScalars.endIndex
+        for offset in 1...string.unicodeScalars.count {
+            let index = string.unicodeScalars.index(endIndex, offsetBy: -offset)
```
norio-nomura: This creates the index from the end of the string every time.
Can you reuse the index created in the previous loop to reduce computational cost?

Lukas-Stuehrk: Yes, you're right. Thanks for the input. I think there's even a much easier implementation (at least for me it's much easier to wrap my head around 😉 ):
```Swift
        var unicodeScalars = self.bridge().unicodeScalars
        while let scalar = unicodeScalars.last {
            if !characterSet.contains(scalar) {
                return String(unicodeScalars)
            }
            unicodeScalars.removeLast()
        }
```

Do you have concerns?

jpsim: Looks fine to me.

norio-nomura: 👍

norio-nomura:
Could you please add a changelog entry? Thanks!

Lukas-Stuehrk:
Sure! Done.

jpsim:
Thanks @Lukas-Stuehrk!

Lukas-Stuehrk:
You're welcome. I'm happy to help on such a handy and important piece in the Swift tooling. Thanks for the fast reviews, good feedback and all your work.

PR #358, opened 2017-04-07
Fix crash in lineAndCharacter(forByteOffset:)

marcelofabri:
See https://github.com/realm/SwiftLint/issues/1006


jpsim:
Awesome, thanks! Could you please add a regression test? i.e. a test that would've previously crashed or returned incorrect data?

jpsim:
Thanks for adding the test! Does it pass on Linux? If so, it'd be great to add it to the `allTests` static.

jpsim:
Sorry for the delay in merging this. Looks great!

PR #359, opened 2017-04-11
Option to include expressions in addition to declarations from sourcekit output

PaulTaykalo:
By this PR, now we can choose which elements should be presented in the resulting SourceKitten output.
This is not a breaking change, by default - comments and declarations will be included in the output.
If `--include-expressions` specified, output will contain expressions as well

Possible usages in documentation:
 - Show hidden dependencies from methods

Other:
 - Find hidden dependencies such as singletons access

### DIFF
--- a/Tests/SourceKittenFrameworkTests/Fixtures/Expression.json
+++ b/Tests/SourceKittenFrameworkTests/Fixtures/Expression.json
@@ -1,0 +1,110 @@
+{
+  "Expression.swift": {
+    "key.diagnostic_stage": "source.diagnostic.stage.swift.parse",
+    "key.substructure": [
+      {
+        "key.filepath": "Expression.swift",
+        "key.parsed_declaration": "class Hidden",
+        "key.annotated_decl": "<Declaration>class Hidden</Declaration>",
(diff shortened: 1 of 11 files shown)


lyricsboy:
This seems like a great feature! Any chance of it getting merged? :)

Tests/SourceKittenFrameworkTests/Fixtures/Expression.json
```diff
@@ -0,0 +1,110 @@
...
+            "key.annotated_decl": "<Declaration>func functionWithHiddenDependency()</Declaration>",
+            "key.fully_annotated_decl": "<decl.function.method.instance><syntaxtype.keyword>func</syntaxtype.keyword> <decl.name>functionWithHiddenDependency</decl.name>()</decl.function.method.instance>",
+            "key.accessibility": "source.lang.swift.accessibility.internal",
+            "key.length": 100,
+            "key.substructure": [
+              {
+                "key.namelength": 6,
+                "key.nameoffset": 100,
+                "key.length": 8,
+                "key.name": "Hidden",
```
lyricsboy: Is this all the detail we can get about this function (which is really an initializer for the `class Hidden`)? It would be great to have a reference of some kind to the declaration of the initializer. Although in this case, that initializer is generated by the compiler and there is no code to reference...

PR #371, opened 2017-05-02
Remove redundant conformance constraint 'U': 'Hashable'

norio-nomura:
Swift 4.0 produces error for this.
```
…/Source/SourceKittenFramework/ClangTranslationUnit.swift:23:33: error: redundant conformance constraint 'U': 'Hashable'
    fileprivate func grouped<U: Hashable>(by transform: (Iterator.Element) -> U) -> [U: [Iterator.Element]] {
                                ^
…/Source/SourceKittenFramework/ClangTranslationUnit.swift:23:85: note: conformance constraint 'U': 'Hashable' inferred from type here
    fileprivate func grouped<U: Hashable>(by transform: (Iterator.Element) -> U) -> [U: [Iterator.Element]] {
                                                                                    ^
```

### DIFF
--- a/Source/SourceKittenFramework/ClangTranslationUnit.swift
+++ b/Source/SourceKittenFramework/ClangTranslationUnit.swift
@@ -20,7 +20,7 @@ extension Sequence where Iterator.Element: Hashable {
 }
 
 extension Sequence {
-    fileprivate func grouped<U: Hashable>(by transform: (Iterator.Element) -> U) -> [U: [Iterator.Element]] {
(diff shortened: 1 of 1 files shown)


norio-nomura:
Oh, I didn't know.
Should we not modify this?

jpsim:
We should still do this to prepare for Swift 4, as I think SR-4731 only tracks continuing to support this in Swift 3 compatibility mode.

norio-nomura:
I see.
Thanks! 🙏

PR #377, opened 2017-05-12
Swift 4.0 on Linux

norio-nomura:
- Add travis job with `swift-4.0-DEVELOPMENT-SNAPSHOT-2017-05-11-a` on Linux
- It can't be built with `swift-4.0-DEVELOPMENT-SNAPSHOT-2017-05-11-a` on macOS since compiler crashes. [SR-4863](https://bugs.swift.org/browse/SR-4863)
- Building with SwiftPM requires Swift 3.1 or later.
  - Building by SwiftPM with Swift 3.0.2 should be able to resurrect if [SR-4848](https://bugs.swift.org/browse/SR-4848) would be fixed.

PR #380, opened 2017-05-12
Chang to explicitly pass `()` to `Result.success()`

norio-nomura:
Is this affected by [SE-0110](https://github.com/apple/swift-evolution/blob/master/proposals/0110-distingish-single-tuple-arg.md)?

Extracted from https://github.com/jpsim/SourceKitten/pull/377 which can be applied without passing test on https://github.com/jpsim/SourceKitten/pull/377.


jpsim:
👍

PR #381, opened 2017-05-12
'subscript' returning `String` will be obsoleted in Swift 4

norio-nomura:
https://github.com/apple/swift-evolution/blob/master/proposals/0163-string-revision-1.md#source-compatibility

Extracted from #377 which can be applied without passing test on #377.


norio-nomura:
🙂

Issue #376, opened 2017-05-10
`swiftDeclaration` not generated for some Objective-C methods.


ainopara:
When using SourceKitten to parse Objective-C headers, I found that some `SourceDeclaration` instance has a nil `swiftDeclaration` property.

After research, I found that is because SourceKit response of the cursorInfo request for those methods does not have a `SwiftDocKey.fullXMLDocs` key.
 (https://github.com/jpsim/SourceKitten/blob/master/Source/SourceKittenFramework/Clang%2BSourceKitten.swift#L212)

The swift interface of those methods can be extracted from `SwiftDocKey.annotatedDeclaration` key in the SourceKit response, so I tried to make a change and it works for me.

I'm not sure if this is a known issue in SourceKit that will be fixed in further release of Swift.

My commit: https://github.com/ainopara/SourceKitten/commit/e88e15370fd1ad67d8e2857da4e266b80e417b6b
My toolchain is from Xcode 8.3.2 (8E2002).

jpsim:
Nice find! Would you be able to make a PR with the same commit from your fork? Thanks!

ainopara:
`XMLElement` currently does not have the `recursiveText` property. I have made a PR there https://github.com/drmohundro/SWXMLHash/pull/131 .

jpsim:
Fixed in #382.

PR #382, opened 2017-05-16
Some method do not have fullXMLDocs, in this case we will trying to extract swiftDeclaration from annotatedDeclaration.

ainopara:
As mentioned in #376, `swiftDeclaration` sometimes return nil.

I think that is due to `key.doc.full_as_xml` is marked as optional in current [SourceKit protocol](https://github.com/apple/swift/blob/swift-DEVELOPMENT-SNAPSHOT-2017-05-15-a/tools/SourceKit/docs/Protocol.md#cursor-info).

`SWXMLHash` is updated becasue `recursiveText` property is first introduced in 3.1.0 version.


norio-nomura:
I opened #383 to fixing build error in `VersionCommand.swift`.

norio-nomura:
Restarted `JOB=Xcode` on travis since #383 has been merged.

ainopara:
Rebased.

jpsim:
Excellent. Thanks!

PR #383, opened 2017-05-16
Add `Version.swift` to `project.pbxproj`

norio-nomura:
Fixes:
```
Source/sourcekitten/VersionCommand.swift:18:15: error: use of unresolved identifier 'Version'
       print(Version.current.value)
             ^~~~~~~
```


jpsim:
Oops. Forgot about this. That why you don't push straight to master! Thanks :)

PR #388, opened 2017-05-21
Fix extensions of nested types

johnfairh:
From realm/jazzy#333.  Given:

```swift
class Outer {
    class Inner {
    }
}

extension Outer.Inner {
}
```

...sourcekitten associates the extension with `Outer` -- ie. `key.name` and `key.usr` match `Outer`.  This happens because of the `source.request.cursorinfo` made for the extension against the `Outer.Inner` name: SourceKit interprets this as a request for info about the `Outer` type and diligently returns it.

This PR handles declaration names containing dots and advances the offset of the cursorinfo to the last portion, which is the actual type of the extension.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@
 
 ##### Bug Fixes
 
-* None.
+* Fix handling of Swift extensions of nested types.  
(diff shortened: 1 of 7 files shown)


jpsim:
Fantastic! 👏

PR #390, opened 2017-05-29
Cache NSString.CacheContainer on Linux

jpsim:
Speeds up SwiftLint linting itself on Linux by ~13x (15s vs 3m23s).


norio-nomura:
Was not it necessary to introduce circular references to the macOS implementation?

PR #393, opened 2017-06-22
Support `swift-4.0-DEVELOPMENT-SNAPSHOT-2017-06-19-a`

norio-nomura:
Tests pass with `swift-4.0-DEVELOPMENT-SNAPSHOT-2017-06-19-a`.
Xcode 9 beta 2 is required to using `swift-4.0-DEVELOPMENT-SNAPSHOT-2017-06-19-a` on macOS.

- Update SWXMLHash to 4.1.0
- `compareJSONString()` supports selecting expected fixture by Swift versions and platfroms
  e.g. fixtures for `Bicycle` are selected from:
  - `LinuxBicycle@swift-4.0.json` … Swift 4.0 or later on Linux
  - `Bicycle@swift-3.2.json` … Swift 3.2 or later
  - `LinuxBicycle.json` … on Linux
  - `Bicycle.json` default
- Requires Swift 3.1 or later


norio-nomura:
> `SourceKitService` bundled with Xcode 9 beta crashes on running tests

http://www.openradar.me/radar?id=4980788026146816

norio-nomura:
>  Maybe just doing a full fixture comparison for the latest macOS Swift version, and a "sanity check" for everything else, would be sufficient?

It is necessary to have a method to correctly check the difference between the results of Linux and macOS. If we can do that, I think we can move to that method. 🤔

jpsim:
We can do that later. This is good to go.

norio-nomura:
Thanks!

PR #397, opened 2017-06-30
Document param/result/discussion for Swift 4

johnfairh:
This adds support for the new format of the *full_as_xml* field in Swift 3.2/Swift 4 -- a `<CommentParts>` container element added.  Causes params/result/etc to be missing from json.
Spotted via realm/jazzy#841.
 
Added new test, updated fixtures for tests (parameters etc. restored).
3.2 stuff done with swift.org 2017-06-25-a.


jpsim:
Sweet, thanks!

PR #416, opened 2017-08-14
Improve accessorUSR(getter:) build time

### LINKED ISSUE #415: Increase/remove -warn-long-function-bodies
`-warn-long-function-bodies=200` is making the build fail on Xcode 9 beta 5 on my local machine (2017 MacBook Pro 15") and on [Travis CI](https://travis-ci.org/realm/SwiftLint/jobs/263931821). Might be worth filing a radar, since it's consistently slower for me in the recent beta.

We should either:
- Increase the limit
- Remove the limit
- Find out why it's taking so long to compile [that function](https://github.com/jpsim/SourceKitten/blob/036cb5042c55eef97cf59ae8670f6d44168fe6bc/Source/SourceKittenFramework/SourceDeclaration.swift#L66-L116) (and fix it).

marcelofabri:
Fix #415

### DIFF
--- a/Source/SourceKittenFramework/SourceDeclaration.swift
+++ b/Source/SourceKittenFramework/SourceDeclaration.swift
@@ -63,6 +63,34 @@ public struct SourceDeclaration {
         return accessorUSR(getter: false)
     }
 
+    private enum AccessorType {
+        case `class`, instance
+
+        var propertyTypeString: String {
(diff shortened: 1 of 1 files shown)


jpsim:
This looks good, but CI is clearly very broken at the moment. 😞 

Thanks!

PR #424, opened 2017-09-20
Add tabWidth parameter (default: 1) for lineAndCharacter

sirlantis:
The line/column information normally takes tab width into account. The `W` in the following example would be reported to be at character 9 (for a tab width of 4), but currently it is reported as 6.

```
Hello\tWorld!
```

I picked 1 as a default value to make this a non-breaking change, but as stated above `4` is probably more reasonable. Let me know how you feel about the change in general and what your preference on the default value is.

P.S. This is in preparation of a PR I wanted to do on SwiftLint, to make the vertical parameter alignment rule handle tabs better.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ##### Enhancements
 
-* None.
+* Add `tabWidth` parameter (default: `1`) for `lineAndCharacter`.  
+  [Marcel Jackwerth](https://github.com/sirlantis)
 
 ##### Bug Fixes
 
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -160,7 +160,7 @@ extension NSString {
             return line.byteRange.location + byteDiff
         }
 
-        func lineAndCharacter(forCharacterOffset offset: Int, tabWidth: Int) -> (line: Int, character: Int)? {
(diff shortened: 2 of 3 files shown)


jpsim:
Is this really what Xcode & LLVM's diagnostic engines do? Honestly don't know, but I'd like to align with whatever they do.

sirlantis:
Good point on calling that out. LLVM doesn't seem to do that:

```
let x = \t\t\ty
// Playground execution failed: error: MyPlayground.playground:1:12: error: use of unresolved identifier 'y'
```

So, a default of 1 certainly would make sense, but maybe I should close this PR and in my PR on SwiftLint use a custom version of `lineAndCharacter` everywhere?

Just for context on what my PR is going to be: the vertical aligned parameters rule doesn't handle tabs correctly, the two examples below both raise a violation, even though the first one is correct with tabWidth=4)

```
foo(bar:2
\tbaz:4)

foo(bar:2
 baz:4)
```

jpsim:
> LLVM doesn't seem to do that:

Thanks for finding that out!

> maybe I should close this PR and in my PR on SwiftLint use a custom version of `lineAndCharacter` everywhere?

Doesn't seem to make sense to have to reimplement this in other libraries. Maybe it'd be better if we made it super clear what this was doing by verbose parameter naming:

```swift
func lineAndCharacter(forCharacterOffset offset: Int, expandingTabsToWidth tabWidth: Int) -> (line: Int, character: Int)? {
  /* ... */
}
```

sirlantis:
Sounds great - will change this tomorrow.

jpsim:
Thanks!

PR #426, opened 2017-10-03
Generate Markdown reference documentation from inline docs

eneko:
## Overview
SourceKitten is a great tool to extract information from a Swift/Objc project. One of the most used commands is `doc`, used to generate structured data from the inline source code documentation.

While there are pretty awesome tools out there, like [Jazzy](https://github.com/realm/jazzy), that do a great job generating documentation from the inline docs (Jazzy uses SourceKitten to extract the inline documentation), the generated output is HTML. This requires for the final reference documentation to be hosted somewhere else outside of the project repository, whether it is GitHub Pages or some other web server.

## Markdown reference documents
This pull request adds a new `mdocs` command to SourceKitten.
```
$ sourcekitten mdocs
```

By running this command, SourceKitten will generate Markdown files for all public types found in the source code, whether these contain inline documentation or not. The list of supported types includes:
- Structs
- Classes
- Enums
- Protocols
- Extensions
- Type aliases

Currently, the documentation output files will be saved to `docs/reference/` relative to where the `sourcekitten mdocs` command is being executed. Documents will be organized by type category (classes, structs, enums...). This output folder might be configurable in the future.

The goal is for the generated Markdown documentation to be committed to the project repo, so it can be easily accessible to users browsing the repository on GitHub, BitBucket, etc. However, one could add `docs` to the `.gitignore` file if there is no desire to commit the documentation files.

The `mdocs` command accepts the exact same parameters as the `doc` command, so it can be used to extract documentation from single Swift files, SPM modules, or Xcode project schemes.

**Example Output**
screenshot 2017-10-03 14 26 06

_Note: This PR requires Swift 4. Backwards compatible support for Swift 3 could be added if requested._


eneko:
Boo

ProbablyRusty:
Curious to know why you closed this, @eneko.

jpsim:
Presumably because I didn't reply within a reasonable time.

To be clear, I think this is awesome and would love to see this revived. I just never had the time to properly code review this.

That being said, there's value in it being a separate tool for now, and I've called it out in notable projects that use SourceKitten in #441.

ProbablyRusty:
I just did a couple quick tests on it, but it was generally working very nicely for me.

PR #274, opened 2016-10-18
[WIP] Introduce SourcekitVariant

norio-nomura:
Add `SourceKitVariant` that replacing `SourceKitRepresentable`.
- Replace complicated `SourceKitRepresentable`'s  operations to simple operations
- No performance penalty
  - Same on whole data parsing
  - Better on partial data parsing

``` swift
func walk1(dictionay: [String: SourceKitRepresentable]) {
    if let _ = dictionay[SwiftDocKey.name.rawValue] as? String {
    }
    if let _ = dictionay[SwiftDocKey.kind.rawValue] as? String {
    }
    if let _ = (dictionay[SwiftDocKey.offset.rawValue] as? Int64).map({Int($0)}) {
    }
    if let _ = (dictionay[SwiftDocKey.length.rawValue] as? Int64).map({Int($0)}) {
    }
    guard let substructures = dictionay[SwiftDocKey.substructure.rawValue] as? [SourceKitRepresentable] else { return }
    for substructure in substructures {
        if let substructure = substructure as? [String: SourceKitRepresentable] {
            walk1(dictionay: substructure)
        }
    }
}
```

will be changed to:

``` swift
func walk2(variant: SourceKitVariant) {
    if let _ = variant[.name]?.string {
    }
    if let _ = variant[.kind]?.string {
    }
    if let _ = variant[.offset]?.int {
    }
    if let _ = variant[.length]?.int {
    }
    guard let substructures = variant[.substructure]?.array else { return }
    for substructure in substructures {
        walk2(variant: substructure)
    }
}
```

PR #428, opened 2017-10-06
Refactor `Request` to make it resilience.

ainopara:
## What does this PR do?
- Make `Request` a struct. It can also be a class, but not enum. Because enum associated value can not have default value. If we want to add a new parameter for a request, we have to break the API. By making `Request` a struct (or class), we can add static function which has the same name and parameters of enum item to keep the expression `Request.editorOpen(file: file).send()` works without change. After this PR, we can add / remove parameters for a request while keep our API unchanged.
- As a result of adding new parameters some new api exposed to public. For example: 
```
public static func editorOpen(
        name: String,
        source: EditorOpen.Source,
        enableSyntaxMap: Bool? = nil,
        enableStructure: Bool? = nil,
        enableDiagnostics: Bool? = nil,
        syntacticOnly: Bool? = nil,
        arguments: [String] = defaultArguments
        ) -> Request
```
- Add `SourceKitDef.swift` which is generated by Sourcery with data from `ProtocolUIDs.def` at `$(swift-root)/tools/SourceKit/include/SourceKit/Core/ProtocolUIDs.def`. 
- Introduce `RequestBuilder` to help `SourceKittenFramework` users to build custom SourceKit request.
- Impliment `EditorOpenInterface`, `ModuleGroups` and `ProtocolVersion` three kinds of request.
- ... and add tests for them.
- At last: Move `overwrite ` to parameter of function `compareJSONString` to make update fixture easier.


jpsim:
Thanks for the PR! I must admit, I'm a bit scared by the sheer scope of these changes though and may not find the courage to review 😅 

image

ainopara:
image
The most majority of these changes is from `SourceKitDef.swift` and `SwiftAssertInterface.json`.

The former is a file generated from `ProtocolUIDs.def` in apple/swift repo (clang preprocessor and Sourcery is used in this process). You can only review the structure of this file.

The latter is just a fixture used by test case `testEditorOpenInterface` which is safe to be ignored in review.

Things need to be reviewd are changes in `Request.swift` and the `RequestBuilder.swift` file which is less than 800 lines of change in total.

Recommended review route is begin with `SourceKitDef.swift` then `RequestBuilder.swift` then `Request.swift` then others.

jpsim:
Thanks for the review tips.

jpsim:
Can you please include the script used to generate `SourceKitDef.swift`? Ideally that'd be a command in the `Makefile`.

ainopara:
Sure.
I'm not sure where to place those template files (including the TypeDef.h copied from SourceKit project). `$(project_root)/Templates` or `$(project_root)/Defines` or any other idea?

ainopara:
It seems Circle CI was trying to build target `sourcekitten` for iPhone Simulator then failed.

ainopara:
Oh, the config file is in version 2 format.
We should move it to `.circleci/config.yml` as https://circleci.com/docs/2.0/migrating-from-1-2/ described.

ainopara:
I realized what's wrong with this PR. It does not contain a config file for Circle CI.
I should rebase on master brunch.

ainopara:
I am not realize #274 has a similar solution until it is closed. I think my current solution for strong typed keys has some room to improve, compareing with solution from @norio-nomura .

I'm planning to seperate this PR to several small PRs. The first part will be introduction of a redesigned `SourceKitDef.swift`. I will make a new pull request once it is ready for review.

I will keep this PR open to track what's left to working on.

PR #432, opened 2017-10-20
Keep user alignment of multi-line Swift declarations

johnfairh:
For something like:
```swift
class C {
    func ff(a: String,
            b: String) {
    }
}
```
...make sure the `key.parsed_declaration` looks like:
```swift
func ff(a: String,
        b: String)
```
...instead of (current):
```swift
func ff(a: String,
            b: String)
```

From eg. realm/jazzy#836.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@
 
 ##### Bug Fixes
 
-* None.
+* Preserve horizontal alignment in multi-line Swift declarations.  
+  [John Fairhurst](https://github.com/johnfairh)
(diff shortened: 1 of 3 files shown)


jpsim:
Nice! I wonder why CI didn't run though 🤔

johnfairh:
CircleCI "enable CI for PRs from forks" setting?

jpsim:
That was it. I've turned it on now. Mind pushing again?

johnfairh:
Ok that looks better!

PR #436, opened 2017-11-09
Support Swift 4.0.2

### LINKED ISSUE #435: Cannot compile on Xcode 9.1
With Swift 4.0.2 in Xcode 9.1, the `characters` property is deprecated. SourceKittenFramework uses this in some places, and since warnings are treated as errors, this causes the build to fail. These property accesses should be removed (which I think is safe to do as long as the code isn't meant to be built with Xcode 8.x).

norio-nomura:
Fixes #435


jpsim:
Could add a change log entry

norio-nomura:
Thanks! 🙏

Issue #384, opened 2017-05-16
`compareJSONString` does not detect changes on Linux


norio-nomura:
It seems comparing nested `NSDictionary` on Linux does not working at all since Swift 3.0.2. 😞 

It works as expected on macOS:
```terminal.sh-session
➜  18:33:37 git:(master) ✗ swift
swift_oss_helper command enabled.
Welcome to Apple Swift version 3.1 (swiftlang-802.0.53 clang-802.0.42). Type :help for assistance.
  1> import Foundation 
  2. NSDictionary(dictionary: ["key.entities": [["key": 0]]]) == NSDictionary(dictionary: ["key.entities": [["key": 1]]])
$R0: Bool = false
  3>  ^D
```

It does not work as expected on Linux.
Swift 3.0.2:
```terminal.sh-session
$ docker run --privileged -it -v `pwd`:`pwd` -w `pwd` --rm norionomura/swift:302 swift -I /usr/lib/swift/clang/include
Welcome to Swift version 3.0.2 (swift-3.0.2-RELEASE). Type :help for assistance.
  1> import Foundation 
  2. NSDictionary(dictionary: ["key.entities": [["key": 0]]]) == NSDictionary(dictionary: ["key.entities": [["key": 1]]])
expression failed to parse, unknown error
  1> import Foundation 
  2. NSDictionary(dictionary: ["key.entities": [["key": 0]]]) == NSDictionary(dictionary: ["key.entities": [["key": 1]]])
Shared Swift state for repl_swift has developed fatal errors and is being discarded.
REPL definitions and persistent names/types will be lost.
$R0: Bool = true
  3> ^D
```
Swift 3.1:
```terminal.sh-session
$ docker run --privileged -it -v `pwd`:`pwd` -w `pwd` --rm norionomura/swift:31 swift -I /usr/lib/swift/clang/include
Welcome to Swift version 3.1 (swift-3.1-RELEASE). Type :help for assistance.
  1> import Foundation 
  2. NSDictionary(dictionary: ["key.entities": [["key": 0]]]) == NSDictionary(dictionary: ["key.entities": [["key": 1]]])
$R0: Bool = true
  3> ^D
```
Swift 3.1.1:
```terminal.sh-session
$ docker run --privileged -it -v `pwd`:`pwd` -w `pwd` --rm norionomura/swift:311 swift -I /usr/lib/swift/clang/include
[...]

PR #439, opened 2017-11-10
Clean up JSON in fixtures

norio-nomura:
Change to use `JSONSerialization.WritingOptions.sortedKeys`, that make it easy to see the differences on future changes and the differences between macOS and Linux.
- Use `JSONSerialization.WritingOptions.sortedKeys` on creating JSON String
- Update JSON fixtures to keys are sorted.

Remove unnecessary fixtures:
- Remove `LinuxSimpleCodeCompletion@swift-4.*.json`, since those are same results with using `-swift-version 3`
- Remove `LinuxBicycle.json`, `LinuxExtension.json` and `LinuxSubscript.json`, since those files are used for Swift 3.1.1 on Linux that can not detect difference because of #384.


jpsim:
Excellent. I didn't know `sortedKeys` was an option now. 👍

PR #440, opened 2017-11-10
Enable `testParseFullXMLDocs()` on Linux

norio-nomura:


jpsim:
Hurray!

norio-nomura:
In Swift 3.1.1 for Linux, the test result can not be trusted because of #384.
Should we write `#if os(Linux) && !swift(>=3.2)` in the tests affected by #384?
Or can we end support to build with Swift 3.1.1?

jpsim:
> Or can we end support to build with Swift 3.1.1?

Was thinking of doing this soon, which is why I just released 0.18.4.

norio-nomura:
Ok. 👍

PR #447, opened 2017-11-22
Deprecate `Request.failableSend()`

norio-nomura:
~~Use `Request.failableSend()` instead of `Request.send()`~~
Change `Request.send()` to `throws`

Some APIs changed to `throws`:
* `File.format(trimmingTrailingWhitespace:useTabs:indentWidth:) throws`
* `Structure.init(file:) throws`
* `SyntaxMap.init(file:) throws`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@
 * SourceKitten now requires Xcode 9 and Swift 3.2+ to build.  
   [Norio Nomura](https://github.com/norio-nomura)
 
+* Deprecated `Request.send()`.  
+  [Norio Nomura](https://github.com/norio-nomura)
+
+* Some APIs changed to `throws`.  
+  * `File.format(trimmingTrailingWhitespace:useTabs:indentWidth:) throws`
+  * `Structure.init(file:) throws`
+  * `SyntaxMap.init(file:) throws`
+  [Norio Nomura](https://github.com/norio-nomura)
+
 ##### Enhancements
 
 * None.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * SourceKitten now requires Xcode 9 and Swift 3.2+ to build.  
   [Norio Nomura](https://github.com/norio-nomura)
 
-* Deprecated `Request.send()`.  
+* Deprecated `Request.send()`. Please use `Request.failableSend()` instead.  
   [Norio Nomura](https://github.com/norio-nomura)
 
 * Some APIs changed to `throws`.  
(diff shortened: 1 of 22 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,15 @@
 * SourceKitten now requires Xcode 9 and Swift 3.2+ to build.  
   [Norio Nomura](https://github.com/norio-nomura)
 
+* Deprecated `Request.send()`.
```
jpsim: "Please use `Request.failableSend()` instead.

Source/SourceKittenFramework/Clang+SourceKitten.swift
```diff
@@ -217,7 +222,9 @@ extension CXCursor {
             return xml.replacingOccurrences(of: " & ", with: " &amp; ")
         }
 
-        let cursorInfo = Request.cursorInfo(file: swiftUUID, offset: usrOffset, arguments: compilerArguments).send()
+        guard let cursorInfo = try? Request.cursorInfo(file: swiftUUID, offset: usrOffset, arguments: compilerArguments).failableSend() else {
+                return nil
```
jpsim: indentation

norio-nomura:
Thanks for reviews! Updated.  ✨

norio-nomura:
@jpsim your CircleCI account seems to be stopped by following error:
> Your 14 day macOS trial has ended. Please go to the macOS plan page and select a plan to continue building.

https://circleci.com/gh/jpsim/SourceKitten/619

jpsim:
CircleCI support granted the jpsim GitHub org with access to the Seed plan in early October, but I've exchanged over a dozen emails with them since that point because this has never been reflected in the CircleCI settings dashboard.

I just got an email from them a few minutes ago asking to check to see if it was working now, so I rebased your last commit and pushed again to see if it would run, and it did.

So we should be all good now.

Thanks for the PR!

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@
 * SourceKitten now requires Xcode 9 and Swift 3.2+ to build.  
   [Norio Nomura](https://github.com/norio-nomura)
 
+* Deprecated `Request.send()`. Please use `Request.failableSend()` instead.
```
jpsim: I feel like we should instead rename `failableSend()` to just `send()`. Throwing APIs in the Swift stdlib aren't prefixed with `failable`. It made sense to call it that when we had both, but now that we're removing that, I don't think it's appropriate anymore.

norio-nomura:
Rebased. ✨

norio-nomura:
Thanks! 🙏

PR #448, opened 2017-11-28
Remove unnecessary Swift 4 conditional compilation

jpsim:


jpsim:
Can't do this because of [SR-6486](https://bugs.swift.org/browse/SR-6486).

PR #449, opened 2017-11-29
Fix generic Swift declaration generation from ObjC

johnfairh:
Spotted by a patient jazzy user in realm/jazzy#910.  When SourceKit produces the full_as_xml as part of a cursorinfo request for a decl *imported from ObjC*, it forgets to XML-escape the ‘Declaration’.  SourceKitten had a fix for ‘&’ but not ‘<’ ‘>’ (generics) - causes truncation of the Swift decl.

This PR changes SourceKitten to always use the annotated_declaration_xml instead of trying to work around the Swift bug -- this XML is generated using different code that has always been escaped properly.  For ObjC, the contents of the two declarations are identical.

Testcase change show Realm ObjC was suffering from this -- various generic declarations missing pieces in Swift.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@
 
 ##### Bug Fixes
 
-* None.
+* Fix Swift declarations when generating Objective-C docs for generic types.  
+  [John Fairhurst](https://github.com/johnfairh)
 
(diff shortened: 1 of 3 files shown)


jpsim:
Excellent fix!

PR #451, opened 2017-11-30
Cache `~/.cocoapods` on CircleCI

### LINKED ISSUE #450: Speed up `pod lib lint` job on CI
Currently takes up the most time by far of all the things we do on CI because we clone all of CocoaPods' Specs repo, which seems extremely inefficient considering that we only rely on two dependencies (SWXMLHash & Yams).

norio-nomura:
Trying to resolve #450 by using [Caching Dependencies](https://circleci.com/docs/2.0/caching/).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,14 @@ jobs:
       - run: make test|xcpretty -r junit
       - store_test_results:
           path: build/reports/
+      - restore_cache:
+          keys:
+            - cocoapods-{{ .Environment.CIRCLE_JOB }}
       - run: pod lib lint
+      - save_cache:
+          key: cocoapods-{{ .Environment.CIRCLE_JOB }}
+          paths:
+            - ~/.cocoapods
 
   swiftpm_4:
     macos:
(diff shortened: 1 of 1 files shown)


norio-nomura:
First invocation: https://circleci.com/workflow-run/25360d9f-5087-4451-9a77-ce006a0a0ec6
screenshot 2017-11-30 11 39 44

Second invocation: https://circleci.com/workflow-run/19bd48b1-f053-4872-9f02-e317ad69d93c
screenshot 2017-11-30 11 40 53

It seems to be worth using cache.

.circleci/config.yml
```diff
@@ -10,7 +10,14 @@ jobs:
       - run: make test|xcpretty -r junit
       - store_test_results:
           path: build/reports/
+      - restore_cache:
+          keys:
+            - cocoapods-{{ .Environment.CIRCLE_JOB }}
       - run: pod lib lint
+      - save_cache:
+          key: cocoapods-{{ .Environment.CIRCLE_JOB }}
```
jpsim: Why does this need to be keyed by the circle job? Why not just `key: cocoapods`?

norio-nomura: I was worried that conflicts might occur if the same key is used for two jobs.
I don't know whether that is required or not.

norio-nomura: I changed the key to `cocoapods`, but it seems to be okay. 👍

jpsim:
Excellent work!

PR #453, opened 2017-11-30
Return `SWIFT_NAME` when generating Objective-C docs.

ulukaya:


jpsim:
Tests are failing and this is missing a changelog entry.

PR #454, opened 2017-12-01
[WIP][Experimental] Use libSwiftPM

norio-nomura:


norio-nomura:
https://bugs.swift.org/browse/SR-6513 does not explain `split()` issue correctly.
I filed https://bugs.swift.org/browse/SR-6523 by myself.

jpsim:
This is interesting to consider, but even when it's made to work, I'm not sure it's worth adding a large external dependency for something that's currently pretty simple (though somewhat fragile).

What do you think @norio-nomura?

norio-nomura:
I also reached the conclusion that using current `libSwiftPM` for `Module.init(spmName:)` is inefficient.
I wanted to make sure that I could actually build `libSwiftPM` as a depndency.

PR #456, opened 2017-12-07
Return `SWIFT_NAME` when generating Objective-C docs.

ulukaya:

PR #458, opened 2017-12-08
Remove surplus entry from `SuperScript.json` added by #456

norio-nomura:


jpsim:
Thanks for catching and fixing.

PR #459, opened 2017-12-08
Improve comparing JSON

norio-nomura:
- Add swift-4.1, 3.3, 4.0.3 and 3.2.3 to `versionedExpectedFilename(for:)`
- Print diff on comparing JSON
- Add `appendFixturesForNewSwiftVersion` operation to `compareJSONString()`  
  This will support appending fixtures for new Swift version.


jpsim:
This is so useful 💯

PR #461, opened 2017-12-15
Prepare for swift 4.1

norio-nomura:
- `JSONSerialization` on Linux escapes solidus like macOS does  
  [SR-6401](https://bugs.swift.org/browse/SR-6401)
- `JSONSerialization` on Linux inserts a space after each key like macOS does  
  [SR-5570](https://bugs.swift.org/browse/SR-5570)


norio-nomura:
This will minimize diffs with fixtures on Swift 4.1 and won't pass tests on Swift 4.1.

PR #462, opened 2017-12-18
Automate Objective-C compiler flag generation

johnfairh:
This adds `sourcekitten doc --objc-xcodebuild --umbrella-header <header> [--module-name]` working in roughly the same way as for Swift, following a hint in the code about `-dry-run`.

The magical part is that `xcodebuild -dry-run` is not a complete no-op: it creates the headermap + vfs files that regularly stymie jazzy users today so we can generate and pass those to clang without having to build the entire project.

Have not squashed commits to maybe help review.  Fixtures change is to modify the Realm header directory structure and accounts for the vast majority of the PR's raw LOC.  Testcase shows the xcodebuild way creates the same json as the clang-direct way.

I expect the clang flag include/exclude list to evolve if this gets used - not claiming to have thought about every single flag, just massaged what the projects I have to hand are using to make them work.

Very open to feedback on CLI design/anything else.


Source/SourceKittenFramework/Xcode.swift
```diff
@@ -103,24 +117,102 @@ private func filter(arguments args: [String]) -> [String] {
...
+    /// Name of the compiler flag used to specify the module name
+    private var moduleFlag: String {
+        switch self {
+        case .swift: return "-module-name "
+        case .objc:  return "-fmodule-name="
+        }
+    }
+
+    /// Regexp pattern matching an invocation of the language's compiler,
+    /// optionally targetting a specific module.
```
jpsim: spelling: `targeting`

jpsim:
Finally someone did this 😄. This has been on my todo list for at least two years, and honestly I don't think I'd ever have gotten to do it. So thanks for your initiative!

At first glance, I'm wondering why an umbrella header must be specified at all, and this doesn't just need to take the module name, parsing the results of the `xcodebuild` command following the matching `-fmodule-name` flag.

Haven't done a thorough review yet.

Tests/SourceKittenFrameworkTests/Fixtures/Realm/Stub.m
```diff
@@ -0,0 +1 @@
+#import <Foundation/Foundation.h>
```
jpsim: Why is this file needed?

Tests/SourceKittenFrameworkTests/Fixtures/Realm@swift-3.2.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "Realm\/RLMArray.h" : {
+    "Realm\/Realm\/RLMArray.h" : {
```
jpsim: This file should be unchanged.

Tests/SourceKittenFrameworkTests/Fixtures/Realm/Stub.m
johnfairh: It's part of the dummy xcodeproject I added to the fixtures to check the Realm headers.  If there are no .m files in the project then there is no objective-C compiling in the xcodebuild output.

Tests/SourceKittenFrameworkTests/Fixtures/Realm@swift-3.2.json
johnfairh: I moved the realm header fixtures down a level to add the xcodeproject files somewhere that looked neat (to me!).

Tests/SourceKittenFrameworkTests/Fixtures/Realm/Stub.m
jpsim: ok

Tests/SourceKittenFrameworkTests/Fixtures/Realm@swift-3.2.json
jpsim: ok, that's fine

johnfairh:
>  why an umbrella header must be specified at all,

Probably betraying my objective-C inexperience:  the 'public' header files only do get mentioned in xcodebuild as part of the CpHeader step (they do not get compiled/given to clang).  I guess technically it would be possible to figure out which module was associated with which CpHeader step via parsing the link command??  hmm.  As long as no-one wants to use jazzy to document their internal 'project' headers.

But that wouldn't give a correct #include order that will definitely compile - this is only specified in the umbrella header.

Or are you saying we should be able to figure out the umbrella header file + its path from the module name?

jpsim:
Sorry for the delay. What I have in mind for this would require a fully compiling Xcode project, but would mirror exactly how this is being done for Swift doc generation at the moment, which is to say it would run `xcodebuild`, find the specified `clang <Module Name>` invocation, and extract the clang arguments from that invocation.

So it wouldn't need to assume anything about the headermap/modulemap or umbrella header, the same header files that contribute to the public API of a framework would be the ones that end up being documented.

johnfairh:
OK.  So for ObjC the equivalent is to run libclang against all the sources in the module and filter out declarations that are not in public headers.  Will give it a go.  Agree will be much better UI.

johnfairh:
Closing this one, may circle back one day.

PR #466, opened 2018-01-05
Add Beak to SourceKitten projects

yonaskolb:
https://github.com/yonaskolb/Beak


jpsim:
Nice project, thanks for adding it here.

PR #473, opened 2018-01-27
Improved searching for libsourcekitdInProc.so on Linux for library_wrapper.swift

abl:
These commits add two new search strategies for Linux to the two existing strategies.

The first one checks to see if swiftenv is configured; if so, it uses `swiftenv prefix` to find the lib directory of the current active Swift environment - which is likely what the user is expecting.

The second one checks to see if `swift` is on `$PATH`; if it is, it's likely that the user is building their own toolchain and that's probably the libsourcekitdInProc.so the user is expecting.

This required a little duplication in SourceKitTests.swift that I'm happy to refactor out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 
 * Return `SWIFT_NAME` when generating Objective-C docs.  
   [Ibrahim Ulukaya](https://github.com/ulukaya)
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the great PR! Sorry it took so long for me to review it.

abl:
Thanks for merging and for being Linux-friendly! :)

PR #477, opened 2018-02-01
Pin Commandant based on revision rather than branch

jpsim:
so that downstream SPM consumers don't get warnings


jpsim:
Now that https://github.com/Carthage/Commandant/releases/tag/0.13.0 has been cut, I'll pin to that rather than a commit hash.

PR #485, opened 2018-02-03
Sort `sourcefiles` on producing `Module.docs`

norio-nomura:
The order of `sourcefiles` are different between SPM and Xcode.
- Sorted alphabetically on SPM
- In the order described in Compile Sources build phase within Xcode project

These affect the differences in fixtures generated by `testCommandantDocs()` and `testCommandantDocsSPM()`.
This change minimizes that difference.


jpsim:
Thanks

PR #486, opened 2018-02-03
Only allow U+000A and U+000D as line break tokens

### LINKED ISSUE #475: Only allow U+000A and U+000D as line break tokens
https://github.com/jpsim/SourceKitten/blob/master/Source/SourceKittenFramework/String%2BSourceKitten.swift#L89

This is probably the cause of https://github.com/realm/SwiftLint/issues/2009#issuecomment-361106713, as when `"u{2028}"` [is used](https://github.com/realm/SwiftLint/issues/2009#issuecomment-361106713) the ranges are incorrect.

That's because that character is a [line separator](https://www.fileformat.info/info/unicode/char/2028/index.htm), but it takes 3 bytes to represent it in utf8:

```
p "\u{2028}".utf8.count
(Int) $R1 = 3
```

marcelofabri:
Fixes #475 

We'll probably need to change this again to fix https://github.com/realm/SwiftLint/issues/1786, but I thought we should do that later.

### DIFF
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -32,6 +32,9 @@ private let commentLinePrefixCharacterSet: CharacterSet = {
     return characterSet
 }()
 
+// https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/LexicalStructure.html#//apple_ref/swift/grammar/line-break
+private let newLinesCharacterSet = CharacterSet(charactersIn: "\u{000A}\u{000D}")
+
 extension NSString {
     /**
     CacheContainer caches:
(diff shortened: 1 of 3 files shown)


Source/SourceKittenFramework/String+SourceKitten.swift
```diff
@@ -32,6 +32,9 @@ private let commentLinePrefixCharacterSet: CharacterSet = {
     return characterSet
 }()
 
+// https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/LexicalStructure.html#//apple_ref/swift/grammar/line-break
```
marcelofabri: CI is failing because this line is too long. Should I remove it or just disable the violation?

jpsim:
Thanks for the fix and sorry for the delay on reviewing this. Merged in 11ca107, 438acd2 & 7c16654.

PR #491, opened 2018-02-11
[SwiftDocKey] Make Substructure type description more explicit

ikesyo:
https://github.com/apple/swift/blob/master/tools/SourceKit/docs/Protocol.md

> <key.substructure>:   (dictionary)  // Contains an array of dictionaries ...


jpsim:
Thanks

PR #492, opened 2018-02-12
Use `swift package show-dependencies` for searching Commandant

norio-nomura:
Since the directory to check out Commandant differs depending on the version of Swift, if you run `swift build` with multiple versions of Swift without `swift package reset`, there may be multiple checked out Commandant directories.
That may cause `testCommandantDocsSPM()` to fail.
This change prevents that failure by getting Commandant's directory used in the test with `swift package show-dependencies`.


norio-nomura:
Thanks! 🙏

PR #493, opened 2018-02-12
[SwiftDocKey] Change `getSubstructure` return type to `[[String: SourceKitRepresentable]]?`

ikesyo:
Ref: #491


jpsim:
I could swear that this didn’t used to work. Anyway, seems to work now. Thanks!

PR #494, opened 2018-02-12
[SwiftDocKey] Add `elements` case

ikesyo:
- A use case: https://github.com/ishkawa/DIKit/blob/4954cc93f304534adaef47de3df812d820e6c95d/Sources/DIGenKit/Extension/Structure%2BProperty.swift#L44-L50
- The definition: https://github.com/apple/swift/blob/e9b9f0f9dfdb61fbefcfe9e09eec3198a236ad08/tools/SourceKit/include/SourceKit/Core/ProtocolUIDs.def#L80

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ##### Enhancements
 
-* None.
+* Add `elements` case to `SwiftDocKey`.  
+  [Sho Ikeda](https://github.com/ikesyo)
 
 ##### Bug Fixes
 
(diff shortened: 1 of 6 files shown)


jpsim:
Cool! Could you add a changelog entry. A small simplest case unit test would also be nice.

ikesyo:
Updated!

jpsim:
Thanks for your contribution and your patience. Because of the merge conflict, I decided to rebase and merge locally. I also applied a small change to actually run the test you added on Linux.

Merged in the following commits: b51d27a, f9b8d79, 7cd235a & 0035dc1.

PR #495, opened 2018-02-17
 Replaced linear line index search with binary search

lvsti:
TL;DR: O(log n) is better than O(n)

Long story:
While profiling Sourcery, I noticed that when processing long files, the code spent a significant amount of time searching for the line index of a given byte/character offset. It turned out that the bottlenecks were these lines:

```swift
let index = lines.index(where: { NSLocationInRange(location, $0.range) })
```

The algorithm behind `index(where:)` is obviously a linear search. But since the `lines` array is already ordered according to the range locations, we can do a binary search instead which performs much better on larger collections.

With this change, processing this [flat.swift.zip](https://github.com/jpsim/SourceKitten/files/1734015/flat.swift.zip) with Sourcery took 37 seconds instead of the original 152. It would be interesting to see the effect on other benchmark projects.

Note: I just dumped the search implementation in a private function, let me know if you want to see it in a separate file/as an extension/whatever.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ##### Enhancements
 
-* None.
+* Replaced linear index search with binary search in NSString extension.  
+  [Tamas Lustyik](https://github.com/lvsti)
 
 ##### Bug Fixes
 
--- a/Source/SourceKittenFramework/String+SourceKitten.swift
+++ b/Source/SourceKittenFramework/String+SourceKitten.swift
@@ -32,6 +32,31 @@ private let commentLinePrefixCharacterSet: CharacterSet = {
     return characterSet
 }()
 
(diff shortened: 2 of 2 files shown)


lvsti:
hello, could anybody perhaps take a look?

jpsim:
Hi, yes! Will take a look. Sorry I'm at over 1,000 GitHub notifications. Hard to find time for open source sometimes 😅

jpsim:
Running SwiftLint on a dozen popular Swift projects with this change yielded performance improvements from 0% to 15% faster. Merged with some minor modifications in 123f972c3042ee346fcd651c36f7a95049376c24, f2f2ee2a3ac448eb303fd260a3a609b5d26b378a, 12ff9b5ae82b9b820920b5294a621d968b85a4a9 & 10aacca845997caf5ce78edbf540ca2e5be6a393.

Thanks!

PR #499, opened 2018-03-08
'flatMap' is deprecated: renamed to 'compactMap(_:)'

norio-nomura:
Should we use `-warnings-as-errors` on Linux CI?


norio-nomura:
> Should we use `-warnings-as-errors` on Linux CI?

Since SwiftPM does not support passing build options to specific target, we need to update all dependencies for silencing warnings. :thinking:

jpsim:
I don't really feel compelled to keep Linux warnings clean for all our transitive dependencies. We can catch these incrementally as we go. Thanks!

Issue #498, opened 2018-03-08
Purpose of String(uidString:)?


ileitch:
[String(uidString:)](https://github.com/jpsim/SourceKitten/blob/master/Source/SourceKittenFramework/Request.swift#L154-L164) is appearing near the top of my profiles, and I'm having a hard time understanding its purpose.

All of `SwiftLangSyntax`, `SwiftDeclarationKind` and `SyntaxKind` have a raw type of `String` therefore their `rawValue` should equal `uidString`. The default just assigns `uidString` to `self` anyway, so I don't see the purpose of first creating the enums?

ileitch:
Ah, it's an optimization: https://github.com/jpsim/SourceKitten/blob/master/Source/SourceKittenFramework/Request.swift#L138

I wonder if this still applies for Swift 4.

jpsim:
It'd be interesting to file a Swift bug on JIRA to see if there's something that can be done on the compiler's side to avoid this penalty.

ileitch:
This needs more investigation when I've got some more time to dig deeper, but initial investigation suggests these String optimizations are no longer needed.

The stack frame selected in blue is a reference point. It's something unrelated to String within SourceKitten that I need to investigate, but here I'm just highlighting it to aid comparison.

My very quick and rough changes are here: https://github.com/jpsim/SourceKitten/compare/master...ileitch:uidstring?expand=1

Before:
<img width="1396" alt="before" src="https://user-images.githubusercontent.com/48235/37236042-112c13bc-2458-11e8-91da-914f17541924.png">

After:
<img width="1396" alt="after" src="https://user-images.githubusercontent.com/48235/37236046-17a93864-2458-11e8-982f-21855cb547d6.png">

As you can see, the `SwiftDocKey.init(rawValue:)` and `String.init(bytes:length:)` frames are no longer present. I did delete `String.init(bytes:length:)` so that's clearly expected, however there's no new frame replacing it above the reference point.

The safety of replacing `init(bytes:length:)` with just `init(cString:)` needs to be investigated further, but it does at least appear that `init(uidString: String)` is no longer relevant in Swift 4.1.

jpsim:
Thanks for looking into it! Mind filing a PR where we can fully assess the impact of changing this?

PR #501, opened 2018-03-10
Remove Swift 3 String optimization.

ileitch:
See #498. This is change is more of a discussion point, needs further investigation.


ileitch:
Regarding `init(bytes:length:)` vs. `init(cString:)` being safe, `String(cString: sourcekitd_uid_get_string_ptr(...))` is also used in here: https://github.com/jpsim/SourceKitten/commit/841efd64dea40485bab899abe58ba521757b4d1d

norio-nomura:
I confirmed that most optimization to avoid `decomposedStringWithCanonicalMapping` is no longer necessary in Swift 4. Thank you for reviewing that optimizations!

PR #503, opened 2018-03-15
Loosen version requirement of Yams in `Package.swift`

norio-nomura:
Since 0.19.1 has `.upToNextMinor(from: "0.5.0")` as requirements, SwiftLint can't update Yams to 0.6.0 without updating SourceKitten.
This change will avoid such situation in future.


jpsim:
Thanks!

PR #506, opened 2018-04-04
Increasing SourceKitten project's Swift version

dirtydanee:
Increased the Swift version in SourceKitten to be 4.1. 
screen shot 2018-04-04 at 19 00 50

Strangely, in the project file it shows 4.0. What is the reason? Does the project file skips the minor and the patch versions?


jpsim:
Thanks for the change!

> setting 4.1 in this setting dropdown sets 4.0 in the pbxproj

https://twitter.com/SmileyKeith/status/979966864666198018

Issue #504, opened 2018-03-28
Include declaration attributes in SourceKitten


dirtydanee:
**When i am running with Xcode 9.2**
```bash
strings /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/sourcekitd.framework/Versions/Current/XPCServices/SourceKitService.xpc/Contents/MacOS/SourceKitService|grep source.decl.attribute
```
**I get the following output:**

```bash
source.decl.attribute.ibaction
source.decl.attribute.iboutlet
source.decl.attribute.ibdesignable
source.decl.attribute.ibinspectable
source.decl.attribute.gkinspectable
source.decl.attribute.objc
source.decl.attribute.objc.name
source.decl.attribute._silgen_name
source.decl.attribute.available
source.decl.attribute.final
source.decl.attribute.required
source.decl.attribute.optional
source.decl.attribute.noreturn
source.decl.attribute._exported
source.decl.attribute.NSCopying
source.decl.attribute.NSManaged
source.decl.attribute.lazy
source.decl.attribute.LLDBDebuggerFunction
source.decl.attribute.UIApplicationMain
source.decl.attribute.unsafe_no_objc_tagged_pointer
source.decl.attribute.inline
source.decl.attribute._semantics
source.decl.attribute.dynamic
source.decl.attribute.infix
source.decl.attribute.prefix
source.decl.attribute.postfix
source.decl.attribute._transparent
source.decl.attribute.requires_stored_property_inits
source.decl.attribute.autoclosure
source.decl.attribute.noescape
source.decl.attribute.nonobjc
source.decl.attribute._fixed_layout
source.decl.attribute._inlineable
source.decl.attribute._specialize
source.decl.attribute.objcMembers
source.decl.attribute.mutating
source.decl.attribute.nonmutating
source.decl.attribute.convenience
source.decl.attribute.override
source.decl.attribute.sil_stored
source.decl.attribute.weak
source.decl.attribute.effects
source.decl.attribute.__objc_bridged
source.decl.attribute.NSApplicationMain
source.decl.attribute.objc_non_lazy_realization
source.decl.attribute.__synthesized_protocol
source.decl.attribute.testable
source.decl.attribute._alignment
source.decl.attribute.rethrows
source.decl.attribute._swift_native_objc_runtime_base
source.decl.attribute.indirect
source.decl.attribute.warn_unqualified_access
source.decl.attribute._cdecl
[...]

PR #507, opened 2018-04-05
Introducing SwiftDeclarationAttributeKind

dirtydanee:
Implementing issue https://github.com/jpsim/SourceKitten/issues/504

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 ##### Enhancements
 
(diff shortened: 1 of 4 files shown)


dirtydanee:
I am running on Swift 4.1, and its visible, they have introduced a bunch of new attributes and removed a few compared to older versions.

What should be the right strategy? I should only be creating attributes that are present in all versions? Or is there a way to somehow define version specific attribute kinds?

jpsim:
Thanks!

PR #509, opened 2018-04-07
Update dependencies

norio-nomura:


norio-nomura:
@jpsim `pod lib lint` is failing since Yams 0.7.0 is not pushed to cocoapods.
I'm not allowed to push new version of Yams.
Could you please do `pod trunk push` on Yams?

jpsim:
Yams 0.7.0 was pushed to CocoaPods in this commit: https://github.com/CocoaPods/Specs/commit/e291153deb8c1c330b0440f947583034ca689fc5

jpsim:
I think the issue is that we run `pod repo update` -> `restore_cache` -> `save_cache`. So restoring the cache always overrides the latest update.

norio-nomura:
Thanks for review! 🙏  I could not notice the overriding.

norio-nomura:
Caching `~/.cocoapods` is broken by this. It would be better to use the method that used on SwiftLint.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jpsim:
> Caching `~/.cocoapods` is broken by this. It would be better to use the method that used on SwiftLint.

I'll do this on a follow-up PR. I don't want to block this one on that, and it works for now, just slower.

PR #510, opened 2018-04-08
Set default Swift version to 4.1 & fix CocoaPods caching on CI

jpsim:


jpsim:
This means we're no longer testing Swift 3.x on CI, which I'm fine with. Once we merge this I'll cut a final release of SourceKitten with Swift 3.x support, then we can drop it on `master`.

PR #511, opened 2018-04-08
Remove all remaining Swift 3.x support

jpsim:


jpsim:
Thanks for fixing this :)

norio-nomura:
For keeping logs of fixtures, removed following:
- Remove Swift 3 fixtures f32fddb
- Revert "Remove Swift 3 fixtures" 945fc26
and force pushed.

PR #512, opened 2018-04-13
Make all SwiftDeclarationAttributeKind cases always available

marcelofabri:
This should make https://github.com/realm/SwiftLint/pull/2152 build on Swift 4.0.


jpsim:
:+1:

PR #514, opened 2018-04-14
Fix a problem locating the source kit shared library on linux.

mhovan:
Lightly tested on recent Ubuntu release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 
 ##### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks for the fix! Could you please use `URL.appendingPathComponent(_:)` instead? https://developer.apple.com/documentation/foundation/url/1780239-appendingpathcomponent

jpsim:
Thanks @mhovan!

PR #522, opened 2018-05-14
Added Weaver to the list of projects with SourceKitten

trupin:
I added Weaver, which is another cool project with code generation features built with SourceKitten. 

Weaver is a declarative, easy-to-use and safe Dependency Injection framework for Swift.
- Declarative because it allows to declare dependencies via annotations directly in the Swift code.
- Easy-to-use because it generates the necessary boilerplate code to inject dependencies into Swift types.
- Safe because it validates the dependency graph at compile time and outputs a nice XCode error when something's wrong.

You can check it out on GitHub: https://github.com/scribd/Weaver

I also wrote this article explaining why I developed it: https://medium.com/scribd-data-science-engineering/weaver-a-painless-dependency-injection-framework-for-swift-7c4afad5ef6a

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ environment variable.
   * https://github.com/yonaskolb/Beak
   * https://github.com/zenzz/vs-swifter-server
   * https://github.com/zenzz/zxxswifter-server
+  * https://github.com/scribd/Weaver
 </details>
 
 ## Complete


jpsim:
Weaver looks super interesting, thanks for adding it to this list.

PR #527, opened 2018-06-06
[CircleCI] Add Swift 4.2 on linux job with skipping failing tests

norio-nomura:
- ~~SourceKit on Swift 4.2 does not generate `SubStructure` and `SyntaxMap` by default.~~
~~It seems to need requesting them by using `key.enablesubstructure` and `key.enablesyntaxmap` explicitly.~~
- ~~will remove WIP when add `swift-4.2-branch` job and change failing tests to skip by `#if swift(>=4.1.50)`~~ done


norio-nomura:
Filed about tests that needs updating: #532, #533

jpsim:
Thanks @norio-nomura!

PR #535, opened 2018-06-11
Enable test parallelization on Xcode 10

norio-nomura:
- Enable test parallelization option of `SourceKittenFrameworkTests` in `sourcekitten` scheme on Xcode 10
- Merge `testSourceKittenFrameworkDocsAreValidJSON` into `testLibraryWrappersAreUpToDate`  
Since both of `ModuleTests.testSourceKittenFrameworkDocsAreValidJSON()` and `SourceKitTests.testLibraryWrappersAreUpToDate()` used same build configuration `sourcekittenXcodebuildArguments`, one of `xcodebuild` instances was failing with error when test parallelization is enabled.
This changes avoid that.


jpsim:
🎉

PR #536, opened 2018-06-15
Added FactoryProvider to the list of projects with SourceKitten

Nonchalant:
I added FactoryProvider, which is another cool project with code generation features built with SourceKitten.

FactoryProvider is Generate boilerplate of factory Swift framework.

You can check it out on GitHub: https://github.com/Nonchalant/FactoryProvider


jpsim:
Thanks!

PR #544, opened 2018-07-27
Update dependencies

norio-nomura:
By this changes, SourceKitten is buildable by SwiftPM with `swift-DEVELOPMENT-SNAPSHOT-2018-07-11-a` or later that removed PackageDescription API v3.


jpsim:
Thanks!

PR #553, opened 2018-08-28
[CircleCI] Avoid auto updating Homebrew on installing `swiftlint`

norio-nomura:
This reduces the `swiftlint lint` step from 35-120 seconds to few seconds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ aliases:
     - checkout
     - run: git submodule update --init --recursive
     - run: swift test
-    - run: brew install https://raw.github.com/Homebrew/homebrew-core/master/Formula/swiftlint.rb && swiftlint lint --strict
+    - run: HOMEBREW_NO_AUTO_UPDATE=1 brew install https://raw.github.com/Homebrew/homebrew-core/master/Formula/swiftlint.rb && swiftlint lint --strict
 
   - &steps-for-linux
     - checkout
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -23,7 +23,7 @@ aliases:
     - checkout
     - run: git submodule update --init --recursive
     - run: swift test
-    - run: brew install swiftlint && swiftlint lint --strict
+    - run: brew install https://raw.github.com/Homebrew/homebrew-core/master/Formula/swiftlint.rb && swiftlint lint --strict
```
keith: If you'd prefer you can instead set `HOMEBREW_NO_AUTO_UPDATE=1`

norio-nomura: That would make it possible to use the old `swiftlint`.

keith: Ah yea, you might want to lock this anyways so an update to an existing rule doesn't break PRs unexpectedly

norio-nomura: Ah, I needed `HOMEBREW_NO_AUTO_UPDATE = 1` for what I intended. Thank you for teaching me.
Duration of `swiftlint` step:
2:00 before: https://circleci.com/gh/jpsim/SourceKitten/3682
6sec after: https://circleci.com/gh/jpsim/SourceKitten/3690

norio-nomura: > Ah yea, you might want to lock this anyways so an update to an existing rule doesn't break PRs unexpectedly

I think that we are expecting to use the latest swiftlint.

jpsim:
Thanks!

PR #555, opened 2018-09-02
Avoid `clean` on running `xcodebuild` if New Build System is used.

norio-nomura:
Since Xcode’s New Build System uses `llbuild`, compiler arguments for `Module` can be retrieved from manifest of `llbuild` instead of parsing `xcodebuild`’s output. Until this changes, the `clean` action was used for `xcodebuild` to ensure compiler arguments, but it is not necessary when using the New Build System.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 ##### Enhancements
 
-* None.
+* If New Build System is enabled on Xcode, the `doc` command does not need to
+  use the `clean` action on `xcodebuild`.  
+  [Norio Nomura](https://github.com/norio-nomura)
 
 ##### Bug Fixes
 
--- a/Source/SourceKittenFramework/ClangTranslationUnit.swift
+++ b/Source/SourceKittenFramework/ClangTranslationUnit.swift
@@ -78,7 +78,7 @@ public struct ClangTranslationUnit {
     - parameter path:                Path to run `xcodebuild` from. Uses current path by default.
     */
     public init?(headerFiles: [String], xcodeBuildArguments: [String], inPath path: String = FileManager.default.currentDirectoryPath) {
-        let xcodeBuildOutput = runXcodeBuild(arguments: xcodeBuildArguments + ["-dry-run"], inPath: path) ?? ""
+        let xcodeBuildOutput = XcodeBuild.cleanBuild(arguments: xcodeBuildArguments + ["-dry-run"], inPath: path) ?? ""
         guard let clangArguments = parseCompilerArguments(xcodebuildOutput: xcodeBuildOutput as NSString, language: .objc, moduleName: nil) else {
             fputs("could not parse compiler arguments\n\(xcodeBuildOutput)\n", stderr)
             return nil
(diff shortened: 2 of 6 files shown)


norio-nomura:
If we still need to build after `-dry-run`, I think that it is better to examine the manifest after building it normally. 🤔

norio-nomura:
> If we still need to build after -dry-run, I think that it is better to examine the manifest after building it normally. 🤔

As below without reverting https://github.com/jpsim/SourceKitten/commit/4f130bec6386958d82abf17c71d8dbceb387f44b:
```patch
diff --git a/Source/SourceKittenFramework/Xcode.swift b/Source/SourceKittenFramework/Xcode.swift
index 78b22ac..b74090d 100644
--- a/Source/SourceKittenFramework/Xcode.swift
+++ b/Source/SourceKittenFramework/Xcode.swift
@@ -215,7 +215,11 @@ ${PROJECT_TEMP_ROOT}
 internal func checkNewBuildSystem(xcodeBuildArguments: [String],
                                   moduleName: String? = nil,
                                   inPath path: String) -> [String]? {
-    // Check `PROJECT_TEMP_ROOT`
+    // generates the manifest file when using the new build system.
+    fputs("Running xcodebuild\n", stderr)
+    _ = XcodeBuild.run(arguments: xcodeBuildArguments, inPath: path)
+
+    // Extract `PROJECT_TEMP_ROOT`
     fputs("Checking xcodebuild -showBuildSettings\n", stderr)
     guard let output = XcodeBuild.run(arguments: xcodeBuildArguments + ["-showBuildSettings"], inPath: path),
         let match = regexForProjectTempRoot.firstMatch(in: output, range: NSRange(output.startIndex..<output.endIndex, in: output)),
```

jpsim:
Yes, you're totally right. 👍

jpsim:
Can you get `PROJECT_TEMP_ROOT` from the xcodebuild log?

norio-nomura:
The minimum xcodebuild log is as follows:
```terminal.sh-session
# Xcode 10
$ xcodebuild -workspace Carthage/Checkouts/Commandant/Commandant.xcworkspace/ -scheme Commandant build -verbose
note: Using new build system
note: Planning build
note: Constructing build description

** BUILD SUCCEEDED ** [0.731 sec]

# Xcode 9
$ DEVELOPER_DIR=/Applications/Xcode.app xcodebuild -workspace Carthage/Checkouts/Commandant/Commandant.xcworkspace/ -scheme Commandant build -verbose
=== BUILD TARGET Result-Mac OF PROJECT Result WITH CONFIGURATION Debug ===

Check dependencies

=== BUILD TARGET Commandant OF PROJECT Commandant WITH CONFIGURATION Debug ===

Check dependencies

** BUILD SUCCEEDED **

```
So using `-showBuildSettings` is required to getting `PROJECT_TEMP_ROOT`.

norio-nomura:
Rewrote and rebased.

norio-nomura:
> projects that don’t use the new build system will now be built twice

That will be the worst case. If the target module is not updated by the first build, second build with `clean` action will be required. In such case, I expect that the first build will be done in a short time.

jpsim:
> In such case, I expect that the first build will be done in a short time.

For CI jobs that run jazzy for example (which relies on this functionality), the first build will build from scratch, then the second clean build will also build from scratch, doubling the time spent building the project.

norio-nomura:
> In that case, it should return after the first build.

Sorry, I should mention that I also added `parseCompilerArguments()` to first build.

jpsim:
Ah I missed that! Let me take another look.

norio-nomura:
Thanks! 🙏

PR #565, opened 2018-11-19
Make SourceKitd initialize public

neakor:
Sometimes SourceKit process actually hangs internally. In this case, the only recovery is to kill the SourceKitService process and restart it. However the current `sourcekitd_initialize` function is `internal`. External projects cannot re-initialize it. This change simply exposes that function.


jpsim:
Thanks for the PR!

The file you modified is code-generated, and CI is failing because your edits don't match the codegen.

Besides, I don't think we should just expose this directly. Perhaps we could wrap it in our own function.

```swift
public struct SourceKitService {
    public func initialize() { ... }
}
```

PR #569, opened 2018-11-25
Add `@dynamicMemberLookup` to `UID`

norio-nomura:
I am not sure if it is worth using this. 🤔


jpsim:
This is pretty nice in some ways, but since there are just a small handful of values here, I’d rather use an enum if anything.

Maybe we could codegen an enum from Swift’s UIDs.py file?

jpsim:
This one: https://github.com/apple/swift/blob/master/utils/gyb_sourcekit_support/UIDs.py

PR #571, opened 2018-11-28
Additional updates related to Swift 4.2

norio-nomura:


Tests/SourceKittenFrameworkTests/DocInfoTests.swift
```diff
@@ -22,9 +22,9 @@ class DocInfoTests: XCTestCase {
     }
 
     func testModuleInfoRequest() throws {
-#if swift(>=4.1.50)
+#if swift(>=4.2)
```
jpsim: Why this change? Same compiler version, no?

Source/SourceKittenFramework/LinuxCompatibility.swift
```diff
@@ -10,50 +10,30 @@ import Foundation
 
 extension Array {
     public func bridge() -> NSArray {
```
jpsim: Should we remove all these bridge functions?

Tests/SourceKittenFrameworkTests/DocInfoTests.swift
norio-nomura: By this PR, SourceKitten will be built with `-swift-version 4.2` swift compiler option.
Under that condition, it will never be `4.1.50`, so I thought that this should be changed.
🤔 
Or should we switch to [compiler version directive](https://github.com/apple/swift-evolution/blob/master/proposals/0212-compiler-version-directive.md) on occasion?

Source/SourceKittenFramework/LinuxCompatibility.swift
norio-nomura: It may be so, but I think it would be better to have another PR.

jpsim: Agreed

Tests/SourceKittenFrameworkTests/DocInfoTests.swift
jpsim: In some cases maybe we should but that’s unnecessary right now.

PR #578, opened 2019-02-23
Improve Module Name Inference in `Module`

chriszielinski:
This PR introduces `XcodeBuildSetting`—a safer, faster alternative to using regex—& improves the module name inference in `Module`. 

The prior solution would fail to build a platform-suffixed target, like `'Result-tvOS'`, due to `moduleName(fromArguments:)`'s assumption that the target/scheme name will be equivalent to the module name. The new implementation will read the module name from the build settings. It will use, in order of priority, the value of the user-defined key `"SWIFT_MODULE_NAME"` & then `"PRODUCT_MODULE_NAME"`. If this fails for whatever reason, it will use `moduleName(fromArguments:)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
   [Minh Nguyễn](https://github.com/1ec5)
   [#574](https://github.com/jpsim/SourceKitten/issues/574)
 
+* Fix `xcodebuild clean` path for new build system and Xcode 10.2.  
+  [John Fairhurst](https://github.com/johnfairh)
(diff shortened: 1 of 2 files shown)


jpsim:
This was a really awesome PR @chriszielinski. Great use of `@dynamicMemberLookup` 👏 

I made a few edits but it somehow closed the PR 😬 

Sorry about that, I'll re-open a new one and post the link here.

jpsim:
Continued in https://github.com/jpsim/SourceKitten/pull/584.

PR #580, opened 2019-03-06
Don't pass swiftc to sourcekit with new build system

johnfairh:
When the new build system `manifest.xcbuild` file is used to figure out compiler args, the path to `swiftc` itself gets included.

This is different to old build system / SPM, and although it is (apparently) ignored by `cursorinfo` it causes problems with other requests (`docinfo` happens to be the one I'm working with).

This PR aligns the new-build-system args finder with the other two kinds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@
   longer prevent documentation generation.  
   [John Fairhurst](https://github.com/johnfairh)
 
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks for the PR! Made a small syntactical change, fixed the changelog conflict and will merge when CI passes.

PR #582, opened 2019-03-20
[Makefile] Avoid using `-Xswiftc -static-stdlib` if toolchain does not have `usr/lib/swift_static` directory

norio-nomura:
This is required on building with Xcode 10.2, because Xcode 10.2’s toolchain does not have `usr/lib/swift_static` directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,10 @@ XCODEFLAGS=-workspace 'SourceKitten.xcworkspace' \
 SWIFT_BUILD_FLAGS=--configuration release
 UNAME=$(shell uname)
 ifeq ($(UNAME), Darwin)
+USE_SWIFT_STATIC_STDLIB:=$(shell test -d $$(dirname $$(xcrun --find swift))/../lib/swift_static/macosx && echo yes)
+ifeq ($(USE_SWIFT_STATIC_STDLIB), yes)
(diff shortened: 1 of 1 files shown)


jpsim:
Thanks for the PR!

Correct me if I'm wrong, but I think this practically means that we should keep deploying binaries with Swift 4.2 because otherwise users will run into issues where the Swift dynamic libraries won't be found on their machines.

norio-nomura:
- `SourceKitten.pkg` for GitHub Release
  [swift-5.0-RELEASE Toolchain](https://swift.org/builds/swift-5.0-release/xcode/swift-5.0-RELEASE/swift-5.0-RELEASE-osx.pkg) distributed from https://swift.org/download/ contains `usr/lib/swift_static` directory. we can use that to packaging `SourceKitten.pkg`.

- Homebrew bottle for macOS 10.14
  After Bottling for macOS 10.14 will be switched to be built with Xcode 10.2, if the user installs it on macOS 10.14 without Xcode 10.2, there will be an issue where Swift dynamic library can not be found.
On such situation, it can be fixed by installing "Swift 5 Runtime Support for Command Line Tools" that can be downloaded from https://developer.apple.com/download/more/.
- Homebrew bottle for macOS 10.13
  Bottle should be built with Xcode 10.1, so no problems occur.

PR #583, opened 2019-03-29
Fix clean xcodebuild with Xcode 10.2 and new build system

johnfairh:
Found via realm/jazzy#1057.

`xcodebuild` in 10.2 requires an additional signing-related flag to skip signing without failing when using the new build system.

The `XcodeBuild.cleanBuild()` path is unreachable on new build system if all is well but we get there if user has wrong settings (etc).

edit: tested this on matrix of Xcode 10.1/10.2, new/old build system, project needs/doesn't need signing.


jpsim:
Thanks for the PR! Fixed the changelog conflict and will merge when CI passes.

PR #585, opened 2019-04-08
Don't document deinit destructors

jpsim:
Fixes https://github.com/realm/SwiftLint/issues/2690


jpsim:
Hmmm, I don't think this worked.

PR #593, opened 2019-05-10
Include parsed extension declarations in docs

johnfairh:
This populates `parsed_declaration` for `extension` declarations.

This is the simplest way I've found to get the actual extension declaration along with any constraints - the cursor-info declarations give us info about the _extended_ type.

Peering back through file history these may have been originally excluded due to overwriting fields from doc-structure with incorrect ones from cursor-info, which was since fixed.  Or there could have been SourceKit bugs now fixed.  At any rate, this version looks to work well for the fixtures here and various jazzy projects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ##### Enhancements
 
-* None.
+* Add parsed extension declarations to Swift docs.  
+  [John Fairhurst](https://github.com/johnfairh)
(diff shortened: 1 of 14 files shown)


jpsim:
I’m glad to see this works now! Thanks for the PR.

There’s a chance this might not work as well with older versions of SourceKit but I’m willing to take that risk.

Issue #595, opened 2019-05-14
Missing ability to query cursorInfo by USR


biboran:
Right now [Request](https://github.com/jpsim/SourceKitten/blob/b039da63355c73c1351b4552cb7818773376534d/Source/SourceKittenFramework/Request.swift#L185) supports retrieving `cursorInfo` only with offset but it can also be useful to fetch `cursorInfo` per [USR](https://github.com/apple/swift/blob/b9877c21f885d1e06cc5d4d7d6653ad033e67380/tools/SourceKit/tools/sourcekitd/lib/API/Requests.cpp#L941).

PR #596, opened 2019-05-14
Add request for fetching cursorInfo from USR

biboran:
As per #595 
* Added a `cursorInfoUSR` request type
* Added a mapping test


jpsim:
This is so great, thanks!

PR #602, opened 2019-07-10
Fix crash with raw strings in Swift 5.0+

jpsim:
Fix crash in `NSString.location(fromByteOffset:)` when run on strings containing raw string literals in Swift 5.0 or later. https://github.com/realm/SwiftLint/issues/2793


jpsim:
Filed a bug against SourceKit: https://bugs.swift.org/browse/SR-11099

Issue #559, opened 2018-10-01
Memory issue with large modules on Linux for Swift 4.2


Fryie:
We have a server-side project (using SPM) and have a script to generate documentation for it. However, when trying to migrate to Swift 4.2 (from 4.1.2), we noticed that in our docker container on CI, the documentation couldn't be generated anymore for one of the modules (the other modules in our project worked fine). Basically, what happened was that the documentation generation started becoming very slow after ca. 30-40 files and then ended up being killed. That problem does not appear on macOS.

This behaviour suggests some kind of memory issue, either a leak, or just an inefficient use of memory (maybe on large modules, some intermediate information could be stored on disk?).

To reproduce this issue I created a [silly repository](https://github.com/Fryie/Sourcekitten-Bug) with a single module that has 2001 source files—all of them very simple, one `Base` class and 2000 children inheriting from it. Our repository has of course fewer files, but they are more complicated; I would suspect that the underlying issue is the same.

Running `sourcekitten doc --spm-module Dummy` on a Linux machine seems to eat up loads of memory. I tested it on a Vagrant machine and sourcekitten was using about 84% of memory while >98% of CPU time was spent swapping. At some point the process is simply killed.

In the linked repository, the issue can be reproduced by running everything inside a container (just by using `make linux42`). The container seems to be able to handle a little bit more than my Vagrant VM does, but at some point it gives up nonetheless. Running htop shows that the memory usage is completely off the charts (it takes less than 5s to grow above 500MB).

This doesn't happen on my mac (just test with `make mac`): While the generation is slow and memory also increases, the progress seems more steady instead of everything becoming progressively slower, and towards the end Activity Monitor shows that < 600MB are being used by sourcekitten, while < 200MB are being used by SourceKit, but this takes a very long time to build up, as opposed to the Linux situation.

We can also test the issue on the Swift 4.1 release (`make linux41`). If I try that, I find out that sourcekitten will be able to process a slightly larger amount of files before grinding to a halt (on my machine, it's about 245 files on Swift 4.2 and about 300 on Swift 4.1 until things start to go bad). This might explain why in our project, it still worked fine with Swift 4.1, but now it doesn't anymore.

Now, of course I'm not sure if I'm on the right track here and/or if this is a sourcekitten issue or a bug in SourceKit.

PR #603, opened 2019-07-17
Refactor SourceKitObject to release memory

cltnschlosser:
I moved all code that interacts with `sourcekitd_object_t` into `SourceKitObject.swift`. This way it's easier to manage the lifecycle of `sourcekitd_object_t` pointers.

Relates to: https://github.com/realm/SwiftLint/issues/2812

I think this could be simplified/improved even more by exposing SourceKitObject as a typealias for `[String: SourceKitObjectConvertible]`. Then when we make the request, converting to `sourcekitd_object_t` pointers, making the request, then freeing the pointers. This means that for the most part you're only working with swift types and the `sourcekitd_object_t` lifecycle is greatly shorted and simplified.

### DIFF
--- a/Source/SourceKittenFramework/SourceKitObject.swift
+++ b/Source/SourceKittenFramework/SourceKitObject.swift
@@ -63,10 +63,10 @@ extension UID: SourceKitObjectConvertible {
 /// Swift representation of sourcekitd_object_t
 public final class SourceKitObject {
     fileprivate let sourcekitdObject: sourcekitd_object_t
-    
+
     /// Other SourceKitObjects whose lifetime is tied to this one (ex: array elements, dictionary values)
     private let children: [SourceKitObject?]
-    
+
     init(yaml: String) {
         self.sourcekitdObject = sourcekitd_request_create_from_yaml(yaml, nil)!
         self.children = []
@@ -76,7 +76,7 @@ public final class SourceKitObject {
         self.sourcekitdObject = sourcekitdObject
         self.children = children
     }
-    
+
     deinit {
         sourcekitd_request_release(sourcekitdObject)
     }
@@ -100,7 +100,7 @@ public final class SourceKitObject {
     public func updateValue<T>(_ value: SourceKitObjectConvertible, forKey key: T) where T: RawRepresentable, T.RawValue == String {
         updateValue(value, forKey: UID(key.rawValue))
     }
-    
+
     /// Swift wrapper for sourcekitd_send_request_sync
     /// Must call sourcekitd_response_dispose on the resulting object.
     func sendSync() -> sourcekitd_response_t? {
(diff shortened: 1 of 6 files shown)


Source/SourceKittenFramework/SourceKitObject.swift
```diff
@@ -81,9 +89,8 @@ public struct SourceKitObject {
     ///   - key: The key to associate with value. If key already exists in the dictionary, 
     ///     value replaces the existing associated value. If key isn't already a key of the dictionary
     public func updateValue(_ value: SourceKitObjectConvertible, forKey key: UID) {
-        precondition(sourcekitdObject != nil)
-        precondition(value.sourcekitdObject != nil)
-        sourcekitd_request_dictionary_set_value(sourcekitdObject!, key.uid, value.sourcekitdObject!)
+        precondition(value.sourceKitObject != nil)
+        sourcekitd_request_dictionary_set_value(sourcekitdObject, key.sourcekitdUID, value.sourceKitObject!.sourcekitdObject)
```
cltnschlosser: This line here worries me slightly because I think we may need to retain `value.sourceKitObject`. Theoretically we could make `children` mutable and just append it to the array. We would be leaking the old value, but only for the lifetime of the dictionary.

My other suggestion in the PR description would also resolve this potential issue.

cltnschlosser:
Looks like I used something exclusive to swift 5, I'll take a look tonight.

cltnschlosser:
@jpsim When you get some time, take a look :)

Source/SourceKittenFramework/SourceKitObject.swift
jpsim: Even with this concern, this is still leaps and bounds better than what we had before. I'm not too concerned in practice.

cltnschlosser: Let me do a little testing. I'm worried that we will set the value and then free it, which could result in memory corruption.

I'll also add those changelogs.

cltnschlosser: There are plenty of unit tests that are stressing this function, but all with `Int64`s. Since they are a c primitive I figured there might be room for this to cause issues in the future. I decided to add this sourceKitObject to the children array, then it's lifecycle will be tied to it's parent. Seems like the safest option.

cltnschlosser:
@fryie Does this solve https://github.com/jpsim/SourceKitten/issues/559, you can wait until after this is merged to test.

Source/SourceKittenFramework/SourceKitObject.swift
jpsim: Sounds good. Thanks for being thorough!

jpsim:
Woohoo, thanks @cltnschlosser!

Fryie:
@cltnschlosser Sorry that I didn't get to it earlier. Unfortunately, this does not appear to fix #559. I updated the [test repository](https://github.com/Fryie/Sourcekitten-Bug) to use the newest version of SourceKitten, but the problem remains.

PR #606, opened 2019-07-22
Fix crash with doc comment on non-declaration

johnfairh:
From realm/jazzy#1083.

#593 relaxed `File.shouldParseDeclaration` to make extension declarations work properly but this regressed a case where there is a (misplaced) doc comment on an expression that `cursorInfo` resolves to a different file at an offset that is invalid for the _current_ file.  We find this doc comment and expression during the `furtherProcess` pass and crash applying the offset to the current file.

This patch filters out non-declarations before doing any processing.


jpsim:
Oops, thanks for the fix. Could we add a test that would have previously crashed?

johnfairh:
OK - the minimal test case is [a couple of source files](https://github.com/realm/jazzy/issues/1083#issuecomment-513868425) -- I'll figure out how to best add that in here.

johnfairh:
Test added - decided to just add the fixtures rather than write them out at runtime.

PR #607, opened 2019-07-24
Bump Commandant to 0.17.0 + Drop Swift 4.2

ultramiraculous:
Bump `Commandant` to 0.17.0 to drop the unnecessary `Result` dependency, and, as an extension, drop Swift 4.2 support.


jpsim:
Thanks for the PR. As far as I can tell, we’ll need to rethink our integration of dependencies with this update, since Commandant no longer uses an Xcode project.

How do you plan on addressing that?

Also, we’ll need to stop importing the Result framework and update a few tests that use Commandant as a fixture.

jpsim:
Closing as inactive. Thanks for the PR nonetheless @ultramiraculous.

PR #611, opened 2019-08-03
Autodetect SPM module name for docs generation

johnfairh:
This adds an `--spm` option to auto-detect a sensible module name.

This is in aid of bringing spm support up to parity with xcodebuild in jazzy -- wanted to get some feedback before going further.

If the idea and this piece look OK then the next is to add a `swift build` call somewhere along the way to make sure the build DB is present and up to date, as we do with xcodebuild.  I think it makes sense to add this call to sourcekitten (vs. jazzy) to preserve the existing layering & keep sourcekitten self-consistent?

Jazzy prototype with this in realm/jazzy#1089

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 ##### Enhancements
 
-* None.
+* Add `--spm` option to guess the name of a Swift Package Manager module
+  for documentation generation.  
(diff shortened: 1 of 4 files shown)


jpsim:
Awesome! Sorry it took me so long to get to this.

PR #613, opened 2019-08-30
Support response files in Swift compiler arguments

johnfairh:
This adds Xcode11 support to docs generation, reading the new response file to figure out the source files.

It turns out SourceKit doesn't support response files ("_filename_ is not part of the input files" error) so we expand them before passing on.

Added a quick test that could go away when the SourceKitten tests pass with Xcode11; jazzy specs all build fine with this change.

### DIFF
--- a/Tests/SourceKittenFrameworkTests/StringTests.swift
+++ b/Tests/SourceKittenFrameworkTests/StringTests.swift
@@ -7,7 +7,7 @@
 //
 
 import Foundation
-import SourceKittenFramework
+@testable import SourceKittenFramework
 import XCTest
 
 class StringTests: XCTestCase {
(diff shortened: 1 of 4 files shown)


Tests/SourceKittenFrameworkTests/StringTests.swift
```diff
@@ -228,6 +228,15 @@ class StringTests: XCTestCase {
         // no crash on bad input
         XCTAssertEqual("ab\\".unescaped, "ab")
     }
+
+    func testResponseFiles() {
+        let responseContents = "3\n4\n5\n"
+        let responseUrl = URL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent("\(UUID().uuidString).rsp")
+        try! responseContents.data(using: .utf8)?.write(to: responseUrl)
```
jpsim: You could mark this test as `throws` and remove the bang

Tests/SourceKittenFrameworkTests/StringTests.swift
```diff
@@ -228,6 +228,15 @@ class StringTests: XCTestCase {
         // no crash on bad input
         XCTAssertEqual("ab\\".unescaped, "ab")
     }
+
+    func testResponseFiles() {
+        let responseContents = "3\n4\n5\n"
```
jpsim: Maybe have a "file" with a space in its path?
```suggestion
        let responseContents = "3 4\n5\n"
```

Might want to separate the test string with some other character than space in that case too (`,` or `|`) for example.

johnfairh:
Thanks for the review, updated the test.  Looks like a maintainer needs to merge though.

jpsim:
Nice. Let me know whenever you need a release too. Is this the last one in your current batch that you want to land or is there more?

johnfairh:
Cheers; #614 should be the final one -- a release after that would be great.

PR #614, opened 2019-09-06
Add package build step to SPM docs generation

johnfairh:
Final part of jazzy+SPM support: this runs `swift build` as part of generating docs for SPM modules.  This makes sure that the build manifest (a) exists and (b) is consistent with the latest changes to the source code, same as Xcode path.  (No need to do a `clean` so is v. cheap if the package has already been built.)

First commit is a pure refactor of all the popen-ish code because I couldn't stand duplicating it again.  Accounts for most of the changed LOC.

Second commit adds a `Module` initializer that runs `swift build` before going to SourceKit.

I tentatively kept the existing SPM initializer to keep semantics for existing users (SourceDocs uses it, didn't look further).

### DIFF
--- a/Source/SourceKittenFramework/Xcode.swift
+++ b/Source/SourceKittenFramework/Xcode.swift
@@ -198,8 +198,7 @@ public func sdkPath() -> String {
     // xcrun does not exist on Linux
     return ""
 #else
-    return Exec.run("/usr/bin/xcrun", "--show-sdk-path", "--sdk", "macosx").string?
-        .replacingOccurrences(of: "\n", with: "") ?? ""
+    return Exec.run("/usr/bin/xcrun", "--show-sdk-path", "--sdk", "macosx").string ?? ""
 #endif
 }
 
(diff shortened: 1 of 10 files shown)


Source/SourceKittenFramework/Xcode.swift
```diff
@@ -50,45 +50,10 @@ internal enum XcodeBuild {
...
-            task.launchPath = pathOfXcodebuild
-            task.currentDirectoryPath = path
-            task.launch()
-        #endif
-        } catch {
-            return Data()
-        }
-
-        let file = pipe.fileHandleForReading
-        defer { file.closeFile() }
```
johnfairh: The [docs](https://developer.apple.com/documentation/foundation/pipe/1414352-filehandleforreading) say this isn't necessary so I left it out.

jpsim:
Looks good! Made some minor edits, will merge and cut a release if CI is happy.

johnfairh:
Great 👍

PR #618, opened 2019-09-20
Fix response file parsing for escaped characters

johnfairh:
The Xcode 11 Swift response files contain backslash escapes for spaces etc.  I'd somehow convinced myself they didn't.

Addresses realm/jazzy#1108.

_Technically_ unescaped spaces are supposed to delimit separate arguments, but Xcode doesn't use this today.  We would need to reimplement [this code](https://github.com/apple/swift-llvm/blob/c5340df2d1aed4cd16fb43e1ce11139a407f2055/lib/Support/CommandLine.cpp#L702) -- I think this version is OK for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@
 
 ##### Bug Fixes
 
-* None.
+* Fix Swift response files for paths including special characters.  
+  [John Fairhurst](https://github.com/johnfairh)
 
 ## 0.25.0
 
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the fix. Since it's just jazzy using this right now that we know of, maybe you could just update the submodule to point to d721f57 and we can defer cutting a release until there's more broadly useful changes.

johnfairh:
OK, will do.

PR #615, opened 2019-09-10
Improve Swift 5.1 compatibility

jpsim:
* Add new libclang types introduced between 0.49 and 0.59.
* Add new `SwiftDeclarationAttributeKind` and `SwiftDeclarationKind` members introduced in Swift 5.1.
* `SyntaxKind` now conforms to `CaseIterable`.
* Update all tests, fixtures, CI infrastructure, and other tooling

PR #619, opened 2019-09-24
Require building with Swift 5.x, support 5.1

larroy:


jpsim:
The Xcode 10.x CI jobs are failing because the Swift compiler can't find SWXMLHash.

```
▸ Compiling Clang+SourceKitten.swift

❌  /Users/vsts/agent/2.155.1/work/1/s/Source/SourceKittenFramework/Clang+SourceKitten.swift:15:8: no such module 'SWXMLHash'

import SWXMLHash
       ^
```

larroy:
Could you give me a hint on how to fix XCode? I don't have experience with the toolchain.

jpsim:
This will fail on Xcode versions lower than 10.2 because SWXMLHash now specifies `SWIFT_VERSION=5` in its Xcode project. I mentioned the same thing in my broader Swift 5.1 PR #615.

> This is work in progress because I think we'll need to drop Swift 4.x to do this due to SWXMLHash specifying `SWIFT_VERSION=5` in its Xcode project.

I'm ok with dropping Swift 4.2 support at this point, but I'd like to do that when also enabling unit tests to pass and run on CI on Xcode 10.3/11.0 as I'm doing in #615.

jpsim:
Thanks for kicking this off @larroy.

PR #622, opened 2019-10-14
Add ByteRange to prevent disambiguation when working with ranges and byte ranges

PaulTaykalo:
## What
Add ByteRange and ByteOffset wrapper to prevent disambiguation between NSRange (byte range ) and NSRange that represents ByteOffset range

## Why
While working with ranges and tokens and string in `Swiftlint` I realized that it is very easy to use one range when you need totally another one. And since NSRange represents both (byte and string,) and those aren't typed, the idea was to use a type that will reprensents values when youre' working in bytes and not-character/strings world.

## Pros
- It is hard to misguide NSRange with ByteRange
- Problems can only happen where conversion is made

## Cons
Breaks backward compatibility

## Additional
All the data that comes from SourceKit can be wrapped with ByteOffset/ByteRange. We can choose at what level we're about to stop.

## Alternatives
User ByteRange<ByteOffset, ByteOffset> types. This probably will do even better.

### DIFF
--- a/Source/SourceKittenFramework/ByteOffset.swift
+++ b/Source/SourceKittenFramework/ByteOffset.swift
@@ -2,7 +2,7 @@
 //  ByteOffset.swift
 //  SourceKitten
 //
-//  Created by JP Simard on 2019-11-02.
+//  Created by Paul Taykalo on 2019-11-02.
 //  Copyright (c) 2019 SourceKitten. All rights reserved.
 //
 
--- a/Source/SourceKittenFramework/ByteOffset.swift
+++ b/Source/SourceKittenFramework/ByteOffset.swift
@@ -1,0 +1,75 @@
+//
+//  ByteOffset.swift
+//  SourceKitten
+//
+//  Created by JP Simard on 2019-11-02.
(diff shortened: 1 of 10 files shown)


jpsim:
Sorry for the delay, I don't have much time to review open source PRs lately.

This is great, I've thought about doing this a few times over the years.

I think we need to keep pushing on consistency though. I've continued this work in this branch: [feature/byte-range-to-prevent-disambiguation](https://github.com/jpsim/SourceKitten/compare/feature/byte-range-to-prevent-disambiguation)

But I think there are even more places that should be returning or taking in `ByteOffset` or `ByteRange`. This would also need a changelog entry. Are you comfortable continuing this work?

PaulTaykalo:
@jpsim Yep, sure. Just let's scope down this PR. I'll take a look at where changes needed to be done.
But do you have in mind any specific places to look/change?

jpsim:
> I mean, What do you think need to be in the final PR as you see it?

I just want to avoid a partial migration where it's not clear what's a byte offset vs a character offset because some APIs have been updated but others haven't.

PaulTaykalo:
Continuing in #639 639

PR #629, opened 2019-11-16
Use StringView instead of excessive caching in NSString

PaulTaykalo:
- Removing cache from NSstring
- The cache was replaced by StringView property

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ##### Breaking
 
-* None.
+* Replace many of the public extensions on String, and its internal
+  caching mechanism `CacheContainer`, with a new `StringView` type.  
+  [Paul Taykalo](https://github.com/PaulTaykalo)
 
 ##### Enhancements
 
--- a/Source/SourceKittenFramework/Clang+SourceKitten.swift
+++ b/Source/SourceKittenFramework/Clang+SourceKitten.swift
@@ -108,7 +108,7 @@ extension CXCursor {
 
     func str() -> String? {
         let cursorExtent = extent()
-        let contents = try! String(contentsOfFile: cursorExtent.start.file, encoding: .utf8)
+        let contents = StringView(try! String(contentsOfFile: cursorExtent.start.file, encoding: .utf8))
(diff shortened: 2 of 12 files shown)


PaulTaykalo:
@jpsim access to byte <-> NSrange without the caching and locks inside of strings.
So, basically, if someone needs to cache the results, the caller is responsible for doing so.

PaulTaykalo:
@jpsim bump

jpsim:
Sorry for the delay, I'm finally looking at this. Have you run preliminary numbers of SwiftLint or other projects that depend on SourceKitten (e.g. jazzy) to see how this affects their performance?

> if someone needs to cache the results, the caller is responsible for doing so.

Do you have an example of what this looks like? How involved is it? For a tool like jazzy, which uses SourceKitten as a CLI executable, it doesn't cache anything, it relies entirely on a single invocation of SourceKitten to generate the documentation data.

---

If we can keep the runtime of jazzy & SwiftLint similar or better to what it was with this excessive caching, I'm very much in favor of removing that caching altogether.

jpsim:
> Do you have an example of what this looks like?

I just found https://github.com/realm/SwiftLint/pull/2964. I'll be looking at that for details.

jpsim:
The impact to jazzy's use case seems unaffected 🕺 

```
$ hyperfine './sourcekitten-master doc --spm-module Yams' './sourcekitten-stringview doc --spm-module Yams' 
Benchmark #1: ./sourcekitten-master doc --spm-module Yams
  Time (mean ± σ):     14.374 s ±  0.691 s    [User: 599.9 ms, System: 120.1 ms]
  Range (min … max):   13.714 s … 15.984 s    10 runs
 
Benchmark #2: ./sourcekitten-stringview doc --spm-module Yams
  Time (mean ± σ):     14.783 s ±  0.651 s    [User: 609.7 ms, System: 116.3 ms]
  Range (min … max):   13.850 s … 15.739 s    10 runs
 
Summary
  './sourcekitten-master doc --spm-module Yams' ran
    1.03 ± 0.07 times faster than './sourcekitten-stringview doc --spm-module Yams'
```

Issue #631, opened 2019-11-22
Add support for code formatting in Obj-C header comments


brycepauken:
This would only require a change in [commentBody()](https://github.com/jpsim/SourceKitten/blob/8c293777af7ed710ecac6f1cae2335cdef2de317/Source/SourceKittenFramework/Clang%2BSourceKitten.swift#L181) in `Clang+SourceKitten.swift` (as best as I can tell!).

Two different scenarios that would be good to support:

    Individual @c Word Highlighting

And:

    Highlighting an entire codeblock:

    @code
    print("Hello world!")
    @endcode

The first should be implementable via regex / wrapping the targeted word in backticks.

The second can be _rougly_ implemented by just replacing `@code` and `@endcode` with triple-backticks, though this would result in different results from Xcode in some scenarios (most of which should be considered error scenarios already / even cause warnings in Xcode, essentially in situations where the keywords are in a different order or count than expected).

brycepauken:
I have a basic implementation of the above strategy locally & will open a PR in a bit, but definitely happy to talk about the change here, including any other ways of tackling this!

jpsim:
Added in #632.

PR #632, opened 2019-11-22
Add support for code formatting in Obj-C header comments

brycepauken:
Possible fix for #631


jpsim:
Thanks for the PR. I'd prefer if we could leverage Clang's existing understanding of this syntax than use a regular expression, however.

jpsim:
Actually, never mind I think this is easiest. Thanks for the fix!

brycepauken:
Agreed with both thoughts here — this should work well for the time being, but leveraging Clang in the future sounds way more ideal. I'll see if I can find a way to feasibly make that work.

Thanks for merging!

PR #634, opened 2019-12-12
Allow inline html tags in ObjC doc comments

### LINKED ISSUE #409: objc doc generation: "fatal error: not text: CXCommentKind(rawValue: 3)"
I encountered this error while trying to generate documentation for an all-objc framework with jazzy.

tl;dr; don't use html tags, use markdown

I don't know if it is worth addressing, but at least the explanation and workaround might be worth this ticket. (It took me some time to understand where the error came from).

The issue error arises when some fields in the documentation contains an html tag. So the workaround is to remove html tags and prefer markdown.

For information here is the header file, command line args and error log of the failing case:
```objc
// my_header.h

/**
 * some foo function
 * @param bar has a <b>purpose</b>
 */
void foo(int bar);
```

```
$> sourcekitten doc --objc my_header.h -- -x objective-c -isysroot "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk -I '' -fmodules
fatal error: not text: CXCommentKind(rawValue: 3): file /[...]/SourceKitten/Source/SourceKittenFramework/Clang+SourceKitten.swift, line 263
```

but the following works fine:
```objc
// my_header.h

/**
 * some foo function
 * @param bar has a **purpose**
 */
void foo(int bar);
```

ultramiraculous:
In the case that there are raw HTML tags in content, don't blow up, just pass along the content. Things like Swift generated headers contain HTML quite a bit and the error is pretty ambiguous when this fatals.

Fixes #409 / realm/jazzy#976

### DIFF
--- a/Source/SourceKittenFramework/Clang+SourceKitten.swift
+++ b/Source/SourceKittenFramework/Clang+SourceKitten.swift
@@ -263,7 +263,8 @@ extension CXComment {
                 let inlineCommand = child.commandName().map { "@" + $0 }
                 return paragraphString + (inlineCommand ?? "")
             }
-            fatalError("not text: \(child.kind())")
+            // Inline child content like `CXComment_HTMLStartTag` can be ignored b/c subsequent children will contain the html.
+            return paragraphString
         }
         return [.para(paragraphString.removingCommonLeadingWhitespaceFromLines(), kindString)]
     }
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the fix!

PR #636, opened 2020-01-07
Update some tests for cursorinfo in Swift 5.1

johnfairh:
Fix up some test cases using `request.cursorinfo` that got broken in 5.1.

CursorInfoUSRTests.testCursorInfoUsrRequest:

`request.cursorinfo` with a USR was substantially downgraded in Swift 5.1 by apple/swift#22825: it only works now if the symbol resolves to a `TypeDecl`-derived type or an `associatedtype` - meaning properties/functions/function-like things fail to resolve.

This appears to be 100% intentional based on the testcase updates in that PR, so I updated this SourceKitten test to use one of the symbols that still resolves on all versions of Swift.

SwiftDocsTests.misc:

For some reason, passing `-sdk` in compiler args is mandatory in Swift 5.1 for `request.cursorinfo` to work at all.  This affected a few testcases that fabricated their own compiler args.  With the fix in, all their 5.1 output reverted to be identical to 5.0 (good news) except for Bicycle which got some 
more info about `NSString` I guess because Foundation is more available.


jpsim:
Thanks!

Issue #633, opened 2019-11-25
Incorrect "key.parsed_declaration" in function with linebreak


minuscorp:
https://github.com/realm/SwiftLint/blob/4aad7e14b477b4f25eb47f7d46577d614ce16924/Source/SwiftLintFramework/Protocols/Rule.swift#L97 gets parsed as 

```swift
func violationRanges(in file: File, kind: KindType,
```
instead of
```swift
func violationRanges(in file: SwiftLintFile, kind: KindType,
                         dictionary: SourceKittenDictionary) -> [NSRange]
```

PR #638, opened 2020-01-09
Improve key.parsed_declaration for multi-line declarations

johnfairh:
Resolves issues including realm/jazzy#896 and looks like sourcekitten #633.

Fix is to use `key.length` to grab text from the file if there is no `key.bodyoffset`.  Complication is that `key.length` is historically buggy; last known bug though fixed via apple/swift#22411 in Swift 5.1 which appears sound, running over the jazzy spec projects and a bunch of others.

The first commit here adds version-checking to SourceKitten.  I've taken a cut-down version of the SwiftLint code and kept it internal as a starting point.  See what you think -- I can see an argument to taking more of the SwiftLint smarts and making it public.

Second commit adds the `key.length` change & a test.  Test fixtures changes reflect things whose declaration stops in the middle of a line -- enum values, here for `Café`:
```diff
- "key.parsed_declaration" : "case Riser, Café, Drop, Bullhorn",
+ "key.parsed_declaration" : "case Riser, Café",
```
...and generic type parameters, here for `S` (not `init` itself):
```diff
- "key.parsed_declaration" : "init<S: Sequence>(_ sequence: S) where ..."
+ "key.parsed_declaration" : "init<S: Sequence"
```
I think this is reasonable.


jpsim:
Thank you so much! This reminds me that @marcelofabri added a SourceKit request over a year ago which landed in Swift 5.1 that allows us to query the Swift compiler version explicitly from SourceKit: https://github.com/apple/swift/pull/22010

We should use that for Swift 5.1 and later.

johnfairh:
Yep, that's the part of SwiftLint I 'borrowed'.

jpsim:
Oh awesome, sorry I commented before reviewing the code changes.

johnfairh:
Awesome, thanks.

PR #639, opened 2020-01-12
Add ByteRange structure to prevent disambiguation when NSRange instead ByteRange is used

PaulTaykalo:
## What
Add ByteRange and ByteOffset wrapper to prevent disambiguation between NSRange (byte range ) and NSRange that represents ByteOffset range

## Why
While working with ranges and tokens and string in `Swiftlint` I realized that it is very easy to use one range when you need totally another one. And since NSRange represents both (byte and string,) and those aren't typed, the idea was to use a type that will reprensents values when youre' working in bytes and not-character/strings world.

## Pros
- It is hard to misguide NSRange with ByteRange
- Problems can only happen where conversion is made

## Cons
Breaks backward compatibility

## Additional
All the data that comes from SourceKit can be wrapped with ByteOffset/ByteRange. We can choose at what level we're about to stop.

## Alternatives
User ByteRange<ByteOffset, ByteOffset> types. This probably will do even better.


jpsim:
Thanks for your continued work on this! I hope you don't mind I pushed a few changes to your branch:

* Added a changelog entry in the "Breaking" section
* I renamed `ByteOffset` to `ByteCount` and used it to describe lengths too, since semantically the same distinction applies. This allowed getting rid of a lot of type conversions.
* I converted a lot more declarations that were using Int/Int64/NSRange values to represent bytes, notably the `Request` members.

CHANGELOG.md
```diff
@@ -2,7 +2,12 @@
 
 ##### Breaking
 
-* None.
+* Replace all uses of `Int`/`Int64`/`NSRange` representing byte offsets
+  to use newly introduced `ByteOffset` and `ByteRange` values instead.
```
PaulTaykalo: ByteCount :)

Source/SourceKittenFramework/SourceLocation.swift
```diff
@@ -13,14 +13,31 @@ import Clang_C
...
+public struct SourceLocation: Comparable {
     public let file: String
     public let line: UInt32
     public let column: UInt32
     public let offset: UInt32
 
-    public func range(toEnd end: SourceLocation) -> NSRange {
-        return NSRange(location: Int(offset), length: Int(end.offset - offset))
+    public func range(toEnd end: SourceLocation) -> ByteRange {
+        return ByteRange(location: ByteCount(Int(offset)), length: ByteCount(Int(end.offset - offset)))
```
PaulTaykalo: ByteCount(Int(offset)) -> ByteCount(offset)

ByteCount can be created with Int64

Source/SourceKittenFramework/StringView.swift
```diff
@@ -125,7 +125,9 @@ public struct StringView {
      Returns a substring from a start and end SourceLocation.
      */
     public func substringWithSourceRange(start: SourceLocation, end: SourceLocation) -> String? {
-        return substringWithByteRange(start: Int(start.offset), length: Int(end.offset - start.offset))
+        let byteRange = ByteRange(location: ByteCount(Int(start.offset)),
+                                  length: ByteCount(Int(end.offset - start.offset)))
```
PaulTaykalo: here as well, we can use `ByteCount(Int(start.offset)) -> ByteCount(start.offset)`

PaulTaykalo:
Few minors with naming and unneeded casting.

Source/SourceKittenFramework/StringView.swift
```diff
@@ -156,13 +158,13 @@ public struct StringView {
 
      - returns: An equivalent `NSRange`.
      */
-    public func byteRangeToNSRange(start: Int, length: Int) -> NSRange? {
```
PaulTaykalo: Also, it seems, that for backward better backward compatibility it is a good ideat o leave

```
byteRangeToNSRange(start: ByteCount, length: ByteCount)
```

Source/SourceKittenFramework/SourceLocation.swift
jpsim: offset in this case is an `Int32` and I didn't think it was worth adding a new initializer for that.

Source/SourceKittenFramework/StringView.swift
jpsim: Int32.

Source/SourceKittenFramework/StringView.swift
jpsim: In this case, since the function name hasn't changed, the compiler error should be enough to let users know how to migrate their code.

PR #651, opened 2020-05-14
Bypass codesign when building Xcode projects

johnfairh:
Continuation of / fixup for #646, fixes realm/jazzy#1183.

If we're going to stop when `xcodebuild` fails we must make sure to set the vars that
disable code signing otherwise it will just fail for signing-related reasons and never
make it to the 'clean build' part.

Build with this change makes the projects from the jazzy issue build cleanly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 ##### Bug Fixes
 
-* None.
+* Always bypass codesigning when building Xcode projects.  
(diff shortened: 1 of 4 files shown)


johnfairh:
Rebased & retested on Xcode 12; @jpsim any opinion on this one?

jpsim (approved):
Sorry for the delay, this looks good to me, thanks for the PR!

johnfairh:
Thanks - no problem - been that kind of a year.

PR #664, opened 2020-08-12
Fix crashes when parsing .m files

jlaws:


jpsim:
Fix looks good. Can you please add a changelog entry?

jlaws:
> Fix looks good. Can you please add a changelog entry?

Done!

PR #669, opened 2020-08-24
Remove Swift 5.0 fixtures or rename to 5.1

jpsim:


jpsim:
Since we dropped support for building and running tests with Swift 5.0 in https://github.com/jpsim/SourceKitten/pull/668

PR #679, opened 2020-11-25
Migrate from Commandant to swift-argument-parser

jpsim:


jpsim:
@johnfairh since jazzy is one of the bigger users of SourceKitten's CLI interface, I want to make sure you get a chance to look at this before I pull the trigger.

Source/sourcekitten/Doc.swift
```diff
@@ -0,0 +1,71 @@
...
+extension SourceKitten {
+    struct Doc: ParsableCommand {
+        static let configuration = CommandConfiguration(abstract: "Print Swift or Objective-C docs as JSON")
+
+        @Option(help: "Only document one file")
+        var singleFile: Bool = false
+        @Option(help: "Name of Swift module to document (can't be used with `--single-file`)")
+        var moduleName: String = ""
+        @Option(help: "Document a Swift Package Manager module")
+        var spm: Bool = false
```
johnfairh, line 14: The `singleFile`, `spm`, and `objc` properties need to be `@Flag` (or something like that, not used the library) otherwise it tries to eat the next token as the value, I think the same problem in other commands (eg. `complete --spm-module`).

Source/sourcekitten/Doc.swift
```diff
@@ -0,0 +1,71 @@
...
+        @Option(help: "Document Objective-C headers instead of Swift code")
+        var objc: Bool = false
+        @Option(help: "Equivalent to --spm --module-name (string)")
+        var spmModule: String = ""
```
johnfairh, line 18: Could take the opportunity of a breaking release to delete the old `--spm-module` -- not a big deal either way!

johnfairh:
Otherwise looks nice.

Source/sourcekitten/Doc.swift
jpsim: Good idea! Done in 032f2e5.

Source/sourcekitten/Doc.swift
jpsim: Good catch! Fixed in 186c4f6.

jpsim:
Thanks for the review @johnfairh! ❤️

PR #684, opened 2020-11-26
Integrate SnapshotTesting library

jpsim:
Update Xcode versions

When generating local fixtures and running on CI.

Only run on the first minor and last minor version of supported Xcodes to minimize the maintenance burden.

E.g. 11.{4,7} and 12.{0,2}


jpsim:
Not sure if this is worth it until SnapshotTesting supports something like cascading snapshot storage https://github.com/pointfreeco/swift-snapshot-testing/issues/413

Maybe we can do it ourselves in the meantime but I haven't done that here yet.

Makefile
```diff
@@ -100,14 +100,12 @@ update_clang_headers:
 update_fixtures: update_fixtures_macos update_fixtures_docker
 
 update_fixtures_macos:
-	for identifier in org.swift.50120190418a org.swift.5120190905a ; do \
-		swift package reset ; \
-		OVERWRITE_FIXTURES=1 xcrun --toolchain $$identifier swift test ; \
+	for xcode in /Applications/Xcode-11.4.1.app /Applications/Xcode-11.7.app /Applications/Xcode-12.app /Applications/Xcode-12.2.app ; do \
```
norio-nomura, line 103: Wouldn't it be better to align Xcode's naming conventions with GitHub Actions and Azure Pipelines?
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#xcode

jpsim: Sure, happy to do that, but this is mostly for maintainers to run locally, not CI. This is how my Xcodes are named so I went with this. As far as I know, you can’t specify Xcode Swift toolchains with their identifiers.

norio-nomura: Hmm🤔 
How about this one?
```suggestion
	shopt -s nullglob; for xcode in /Applications/Xcode*{11.4.1,11.7,12,12.2}.app ; do \
```

jpsim: Great suggestion, thanks!

PR #692, opened 2021-03-16
Include Glibc to support stderr.

justinctlam:
Include Glibc to support stderr in order to compile successfully in a Linux environment.


jpsim:
Thanks! I wonder why our Linux CI didn't catch this.

PR #698, opened 2021-05-29
Swift 5.4

johnfairh:
Attempt at CI on Swift 5.4 / Xcode 12.5.


johnfairh:
I forgot how the Linux fixtures work, will update.  

I _think_ the library wrappers need to be tested only on 5.4 otherwise the s/UInt8/CChar fails - is harmless.

Worst, GitHub Actions somehow doesn't support macOS 11 / Xcode 12.5 any more -- my personal account got grandfathered in so I'd been blissfully unaware of the mess.  For SourceKitten I think it should be OK to just install a 5.4 toolchain on macOS 10 --- not sure about the libclang equivalent, may have to leave that until GH sort themselves out.

jpsim:
Thanks for taking this on! Do you need me to look into how to get CI running and passing with Swift 5.4?

johnfairh:
The only bits left are the tests that need `libclang.dylib`: it's not in the swift toolchain package, llvm.org latest is too old, and I can't see how to get hold of the (apple internal?) version Xcode uses.  So I'm thinking about just skipping those tests in CI for now -- do you have any better ideas?

johnfairh:
Ok so even with libclang ignored, it looks like the swift.org 5.4 toolchain is too different to whatever ships in Xcode 12.5 -- failure is because what Xcode 12.5 spells as `[Array<String>.Subsequence]`, 5.4 spells as `[ArraySlice<String>]`.

jpsim:
I just got access to GitHub Actions `macos-11` images on the jpsim GitHub organization, can you try using that with Xcode 12.5 directly? It worked here: https://github.com/jpsim/Yams/pull/323

johnfairh:
Oh that's great -- this was all getting a bit torrid.  Ready to squash + merge I think.

(just in time for macOS 12)

jpsim (approved):
Some neat changes in the json fixtures, no regressions from Swift 5.3, just enhancements.

`key.effective_access` is an interesting new key in the index response.

jpsim:
Thanks for shepherding this through @johnfairh ! 👏

PR #699, opened 2021-06-02
Add ability to specify which files should be included and excluded when documenting a module.

mrabiciu:
This change allows us to pass in a list of files to explicitly include and explicitly exclude from the doc command. This can dramatically improve parsing performance for large projects where we only need to document a subset of the files. 

This change is specifically targeted at jazzy's use case and is meant to mirror the behavior of the `include` and `exclude` flags from that package. Currently the entire module is parsed even if jazzy is given `include` and `exclude` flags. With this change jazzy should be able to forward these flags to `sourcekitten`

In our case this change was able to reduce the parsing time by 80% when generating docs, going from 15 minutes to 3 minutes.

### DIFF
--- a/Source/SourceKittenFramework/Module.swift
+++ b/Source/SourceKittenFramework/Module.swift
@@ -35,8 +35,10 @@ public struct Module {
                           package if `nil`.
      - parameter path:    Path of the directory containing the SPM `.build` directory.
                           Uses the current directory by default.
+     - parameter include: Source file pathnames to be included in documentation. Supports wildcards. Empty array will include all files.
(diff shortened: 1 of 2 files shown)


johnfairh:
This looks like a good idea, have had a few jazzy users wanting this kind of thing.

Could do with some unit tests for those extensions. `NSPredicate` is [incomplete on non-Darwin](https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/NSPredicate.swift) so I think that part needs a different approach.

What should we do about objc mode?

See `CONTRIBUTING.md` for changelog advice.

PR #702, opened 2021-06-21
Fix build errors with Xcode 13 beta

thii:


thii:
@jpsim Can you review?

DesmanLead:
+1 for reviewing this, this is the only thing that breaks our build on Xcode 13.

PaulTaykalo:
It Seems that the `master` branch is working fine in Xcode 13 beta 4.
What's the usecase? How are you trying to compile Sourcekitten?

johnfairh:
It was broken in earlier betas, some Swift 5.5 regression that's fixed in latest.

PaulTaykalo:
Will close then PR then. leave a comment if it's still a thing

PR #712, opened 2021-10-05
Add sourcekitten completion results formatting options

### LINKED ISSUE #710: Ability to turn off JSON pretty printing and sorting of keys for better performance
SourceKitten can be quite slow in printing large number of code completion results. For example when importing Foundation:

`sourcekitten complete --text $'import Foundation\n' --offset 18 -- -sdk macosx`

On my machine, this example takes about 1.7 seconds. I've discovered that most of this is due to pretty printing and sorting of the JSON results, which is arguably unneccessary in most cases.

Rebuilding sourcekitten without the `[.prettyPrinted, .sortedKeys]` options in JSONOutput.swift reduces the runtime of the above command to 0.92 seconds, a significant improvement.

Please add an option to disable these printing modes. Thanks!

PaulTaykalo:
Since JSON-formatting of code completion results can take a lot of time in some cases, additional format options can be passed to the sourcekitten.

Two additional options added:
- `prettified` - whether JSON results should be user-friendly prettified
- `sorted-keys` - whether the keys in JSON should be sorted keys

This fixes #710

### DIFF
--- a/Source/SourceKittenFramework/JSONOutput.swift
+++ b/Source/SourceKittenFramework/JSONOutput.swift
@@ -4,24 +4,29 @@ import Foundation
  JSON Object to JSON String.
 
  - parameter object: Object to convert to JSON.
+ - parameter options: Serialization options
 
  - returns: JSON string representation of the input object.
  */
-public func toJSON(_ object: Any) -> String {
+public func toJSON(_ object: Any, options: JSONSerialization.WritingOptions? = nil) -> String {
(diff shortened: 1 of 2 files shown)


PaulTaykalo:
This is open question whether to make these options available at all, or what options should be by default. If you have some other thoughts on that I would gladly rewrite it :) or even close :)

jpsim (approved):
Looks good, thanks!

PR #724, opened 2022-01-24
Bump minimum Swift version to 5.3 & update dependencies

jpsim:


jpsim:
Requires bumping the minimum Swift version to 5.3. I'll do that after cutting a release.

PR #727, opened 2022-02-02
Fix docs generation with Swift 5.6

johnfairh:
Swift 5.6 SourceKit has changed the `source.lang.swift.decl.enumcase` attributes in a way that crashes `sourcekitten doc` on any code with enum cases.

Recall that `enumcase` is the wrapper around a set of `enumelement`s with actual info.  In earlier Swift, `enumcase` had `key.nameoffset` and `key.namelength` fields but these were always 0.  In Swift 5.6 they're entirely gone.

A net gain for sourcekitten because it turns out we were doing useless cursorinfo queries.

This patch is enough to make jazzy work normally, not found any other breakage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@
 
 ##### Enhancements
 
-* None.
+* Support docs generation on Swift 5.6.  
+  [John Fairhurst](https://github.com/johnfairh)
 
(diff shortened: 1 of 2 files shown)


jpsim (approved):
Thanks John!

PR #737, opened 2022-06-10
Filter out unsupported Swift 5.7 flags

johnfairh:
Docs generation fails at cursor-info on Xcode 14 because of a new flag:

Screenshot 2022-06-10 at 12 40 52

It turns out that sourcekitten is already filtering out a suspiciously-similarly named flag so I've just added this new one.  Jazzy etc. work fine with this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,11 @@
 
 ##### Enhancements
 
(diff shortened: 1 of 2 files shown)


jpsim (approved):
Thanks! Sorry for the delay in reviewing.

PR #763, opened 2023-03-12
Resolve tilde when asking for absolute path

SimplyDanny:
This resolves https://github.com/realm/SwiftLint/issues/3554.


jpsim (approved):
Thanks!

PR #766, opened 2023-03-18
Mildm8nnered fix crash on non utf8 files

### LINKED ISSUE #765: `File.contents` will crash for files that cannot be read later, when initialised with `init(pathDeferringReading:)`
This can be seen by running SwiftLint against swift

```
brew install swiftlint
git clone https://github.com/apple/swift
cd swift
swiftlint
```

The problem file is `./validation-test/IDE/crashers_fixed/033-swift-identifier-isoperatorslow.swift `, which cannot be interpreted as UTF-8

```
% swiftlint  ./validation-test/IDE/crashers_fixed/033-swift-identifier-isoperatorslow.swift          
Linting Swift files at paths ./validation-test/IDE/crashers_fixed/033-swift-identifier-isoperatorslow.swift
Linting '033-swift-identifier-isoperatorslow.swift' (1/1)
SourceKittenFramework/File.swift:20: Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=NSCocoaErrorDomain Code=261 "The file “033-swift-identifier-isoperatorslow.swift” couldn’t be opened using text encoding Unicode (UTF-8)." UserInfo={NSFilePath=/Users/martin.redington/Documents/Hudl/Source/swift/validation-test/IDE/crashers_fixed/033-swift-identifier-isoperatorslow.swift, NSStringEncoding=4}
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace.
zsh: trace trap  ../Sportscode64/Pods/SwiftLint/swiftlint 
```

See https://github.com/realm/SwiftLint/issues/4821

But actually, if the file cannot be read for any reason, then this will cause a crash

```
touch t.swift
chmod -r t.swift
swiftlint  t.swift
Linting Swift files at paths t.swift
Linting 't.swift' (1/1)
SourceKittenFramework/File.swift:20: Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=NSCocoaErrorDomain Code=257 "The file “t.swift” couldn’t be opened because you don’t have permission to view it." UserInfo={NSFilePath=/Users/martin.redington/Documents/Hudl/Source/SourceKitten/t.swift, NSUnderlyingError=0x600002a48210 {Error Domain=NSPOSIXErrorDomain Code=13 "Permission denied"}}
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace.
zsh: trace trap  ../Sportscode64/Pods/SwiftLint/swiftlint t.swift
```

mildm8nnered:
Resolves #765

Any `File`'s initialised with `init(pathDeferringReading:)` would then crash on `_contents = try String(contentsOfFile: path!, encoding: .utf8)` if they could not be read as `.utf8`, or actually for any other reason. For example.

```
touch t.swift
chmod -r t.swift
swiftlint  t.swift
Linting Swift files at paths t.swift
Linting 't.swift' (1/1)
SourceKittenFramework/File.swift:20: Fatal error: 'try!' expression unexpectedly raised an error: Error Domain=NSCocoaErrorDomain Code=257 "The file “t.swift” couldn’t be opened because you don’t have permission to view it." UserInfo={NSFilePath=/Users/martin.redington/Documents/Hudl/Source/SourceKitten/t.swift, NSUnderlyingError=0x600002a48210 {Error Domain=NSPOSIXErrorDomain Code=13 "Permission denied"}}
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace.
zsh: trace trap  ../Sportscode64/Pods/SwiftLint/swiftlint t.swift
```

We now catch the error, and set `_contents` to nil.

The error messaging here, and in the failable initialiser could be improved to cover the common cases - file does not exist, file is unreadable, file is not readable as utf-8, but I didn't do that here.


mildm8nnered:
> What would you think of making a throwing accessor instead? Right now programmatically you can't tell if the file was actually empty or it failed to be decoded.

So that would be much nicer - I was thinking of inspecting the exceptions and providing better error messaging, but if the client can just catch and inspect the error themselves, that would be much better.

```
    public var contents: String {
        get throws {
            try _contentsQueue.sync {
                if _contents == nil {
                    _contents = try String(contentsOfFile: path!, encoding: .utf8)
                }
            }
            return _contents!
        }
        set {
            _contentsQueue.sync {
                _contents = newValue
                _stringViewQueue.sync {
                    _stringView = nil
                }
            }
        }
    }
```

Gives me `'set' accessor is not allowed on property with 'get' accessor that is 'async' or 'throws'`

Assuming we could remove the `set` (which is not apparently used within SourceKitten or its unit tests, but is used by SwiftLint in a couple of places), then there are almost 20 callsites for `contents` internally.

Writing `(try? contents) ?? "")` repeatedly was annoying so I ended up with 

```
    internal var safeContents: String {
        (try? contents) ?? ""
    }
```

Not fond of the naming there. That seems nicer for the external caller (unless they hate catching exceptions), but it's a bit annoying to have to touch so many call sites.

https://github.com/jpsim/SourceKitten/compare/main...mildm8nnered:mildm8nnered-fix-crash-on-non-utf8-files-throwable-accessor?expand=1

I think there is a way to arrange the API so that there is

1) a way to access contents that will throw for bad files
2) a way to set the contents (via a method perhaps, if we can't have a throwable property)
3) a safe easy way to access the property without having to deal with the error internally

But the current API `contents` property API is so easy to use right now, and this is a relatively rare case, so I think the current behaviour, where SourceKitten just logs it is not too awful (although that message could be a bit better).

Unless I've missed something and there is an elegant way to handle this?

mildm8nnered:
If we don't end up throwing the error, I was thinking of looking for these specific exceptions and providing slightly better descriptions for them, and then a generic message for any other error.

```
Error Domain=NSCocoaErrorDomain Code=261 "The file “033-swift-identifier-isoperatorslow.swift” couldn’t be opened using text encoding Unicode (UTF-8)." UserInfo={NSFilePath=/Users/martin.redington/Documents/Hudl/Source/swift/validation-test/IDE/crashers_fixed/033-swift-identifier-isoperatorslow.swift, NSStringEncoding=4}

Error Domain=NSCocoaErrorDomain Code=260 "The file “NonUTF8File” couldn’t be opened because there is no such file." UserInfo={NSFilePath=/Users/martin.redington/Documents/Hudl/Source/SourceKitten/Tests/SourceKittenFrameworkTests/Fixtures/NonUTF8File, NSUnderlyingError=0x600000c5c330 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
```

jpsim (approved):
Sure, let's go with this approach. As you mentioned, it's not a common case.

jpsim:
Thanks for doing this.

PR #769, opened 2023-05-11
Windows Support

compnerd:
This initial patch series beings the work to support Windows.  It is incomplete in the sense that there are multiple issues with the tests that need to be ironed out still.  However, the core of the changes seem correct to partially work through the test suite.

Of note is that the implementation here significantly diverges from Linux in at least one one sense: it is generally far more approximate to Darwin as the Windows Swift toolchain is a complete distribution and that includes libclang.  This toolchain is also (partially) able to host the ObjC paths though the Foundation headers are not available as that is not part of the toolchain distribution.

One of the biggest issues that currently need to be resolved in this core set of changes is the toolchain installation location.

If it is helpful/reasonable, it should be possible to peel off parts of this series into individual PRs to get them merged earlier rather than later to help get more things in place (and make it easier for others to contribute as well).


jpsim:
Hi @compnerd, first sorry for the delay (laid off, job search, other priorities, etc).

Second, it's fantastic to see SourceKitten working on Windows, massive kudos 🙌.

My main point of feedback is that some changes break the non-Windows build or functionality. For example, some of the hardcoded paths (e.g. `C:\Program Files\Git\cmd\git.exe`, `.build\checkouts\`, `C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swift.exe`) apply unconditionally on all OSes, where they should only be applied to Windows.

If you don't mind, I'll cherry pick some of your commits (keeping you as author) that don't break anything in the currently supported platforms to reduce the scope of the changes you want to land in this PR. As those smaller pieces get merged in, you can rebase this one as you fix the build/functionality regressions on macOS/Linux.

compnerd:
> Hi @compnerd, first sorry for the delay (laid off, job search, other priorities, etc).

Thank you for the response!  Life often gets in the way, I hope that everything has worked out well now.

> Second, it's fantastic to see SourceKitten working on Windows, massive kudos 🙌.

Thank you!

> My main point of feedback is that some changes break the non-Windows build or functionality. For example, some of the hardcoded paths (e.g. `C:\Program Files\Git\cmd\git.exe`, `.build\checkouts\`, `C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swift.exe`) apply unconditionally on all OSes, where they should only be applied to Windows.

Absolutely on point.  In fact, they are being broken even further as we speak :). They are not entirely portable, and I've tried to call them out in the commits with the "HACK" prefix.  They are not ready at all for merging and actually need some guidance on how to best approach the problems there.

> If you don't mind, I'll cherry pick some of your commits (keeping you as author) that don't break anything in the currently supported platforms to reduce the scope of the changes you want to land in this PR. As those smaller pieces get merged in, you can rebase this one as you fix the build/functionality regressions on macOS/Linux.

That sounds great!  The changes are ordered such that the ones marked as HACK are situated on top to help splitting things that are ready from those that are not.

jpsim:
I've cherry-picked and merged many of the commits from this PR in #779.

Could you please rebase this PR with the remaining changes and I can add feedback inline?

Or you can open new PRs for incremental changes as they're ready if you prefer, up to you.

Tests/SourceKittenFrameworkTests/SourceKitTests.swift
```diff
@@ -20,6 +20,210 @@ private let sourcekitStrings: [String] = {
         }
         fatalError("Could not find or load libsourcekitdInProc.so")
     }()
+#elseif os(Windows)
```
jpsim, line 23: This is now defined in `SourceKitStrings+Windows.swift` as of #779.

compnerd:
Rebased the changes.  I think that we need to add a helper to locate the executable (basically implement a poor `which`).  The location of `swiftc.exe` is going to change soon as well.  This will break this and will need to be aware of looking up the paths by searching `Path`.

Tests/SourceKittenFrameworkTests/ModuleTests.swift
```diff
@@ -10,7 +10,7 @@ let projectRoot: String = bazelProjectRoot ?? #file.bridge()
 
 class ModuleTests: XCTestCase {
 
-#if !os(Linux)
+#if !(os(Linux) || os(Windows))
```
jpsim, line 13: Since Xcode is only available on macOS, can you please change this to `#if os(macOS)`? I probably should have done that initially too.

compnerd: Done with #780.

Tests/SourceKittenFrameworkTests/ModuleTests.swift
```diff
@@ -22,24 +22,24 @@ class ModuleTests: XCTestCase {
         let temporaryURL = URL(fileURLWithPath: NSTemporaryDirectory())
             .appendingPathComponent("\(#function)-\(NSUUID())")
         try FileManager.default.createDirectory(at: temporaryURL, withIntermediateDirectories: true)
-        let cloneArguments = ["git", "clone", "https://github.com/Carthage/Commandant.git"]
-        let cloneResult = Exec.run("/usr/bin/env", cloneArguments, currentDirectory: temporaryURL.path)
+        let cloneArguments = ["clone", "https://github.com/Carthage/Commandant.git"]
+        let cloneResult = Exec.run("C:\\Program Files\\Git\\cmd\\git.exe", cloneArguments, currentDirectory: temporaryURL.path)
```
jpsim, line 26: Can you introduce new functions in `Exec.swift` that abstracts platform-specific execution of the commands we use in SourceKitten? For example:

```swift
extension Exec {
    static func runGit(_ arguments: String...,
                       currentDirectory: String = FileManager.default.currentDirectoryPath,
                       stderr: Stderr = .inherit) -> Results {
#if os(Windows)
        let command = #"C:\Program Files\Git\cmd\git.exe"#
#else
        let command = "/usr/bin/env"
        let arguments = ["git"] + arguments
#endif
        return run(command, arguments, currentDirectory: currentDirectory, stderr: stderr)
    }
}
```

Source/SourceKittenFramework/Module.swift
```diff
@@ -50,7 +50,7 @@ public struct Module {
                 return nodeModuleName == spmName
             }
             let inputs = node["inputs"]?.array(of: String.self) ?? []
-            return inputs.allSatisfy({ !$0.contains(".build/checkouts/") }) && !nodeModuleName.hasSuffix("Tests")
+            return inputs.allSatisfy({ !$0.contains(".build\\checkouts\\") }) && !nodeModuleName.hasSuffix("Tests")
```
jpsim, line 53:
```suggestion
#if os(Windows)
            let checkoutsPath = #".build\checkouts\"#
#else
            let checkoutsPath = ".build/checkouts/"
#endif
            return inputs.allSatisfy({ !$0.contains(checkoutsPath) }) && !nodeModuleName.hasSuffix("Tests")
```

Source/SourceKittenFramework/library_wrapper.swift
```diff
@@ -1,16 +1,62 @@
 import Foundation
+#if os(Windows)
+import WinSDK
+
+@_transparent
+internal func MAKELANGID(_ p: WORD, _ s: WORD) -> DWORD {
+    return DWORD((s << 10) | p)
+}
+
+struct WindowsError {
```
jpsim, line 10: Can you please move this to its own file? `WindowsError.swift`

jpsim: Should this conform to `Error`?

Tests/SourceKittenFrameworkTests/ClangTranslationUnitTests.swift
```diff
@@ -30,8 +30,14 @@ class ClangTranslationUnitTests: XCTestCase {
     }
 
     private func compare(clangFixture fixture: String) {
-        let unit = ClangTranslationUnit(headerFiles: [fixturesDirectory + fixture + ".h"],
-                                        compilerArguments: ["-x", "objective-c", "-isysroot", sdkPath(), "-I", fixturesDirectory])
+        let path = URL(fileURLWithPath: fixturesDirectory + fixture + ".h")
```
jpsim, line 33: This pattern is used a lot now: ` URL(fileURLWithPath: ...).standardizedFileURL.withUnsafeFileSystemRepresentation { String(cString: $0!) }`

Can you add some sort of shared helper for this? Open to API design ideas, but as a strawman: `FileUtils.unsafeSystemPath(for path: String) -> String`

Tests/SourceKittenFrameworkTests/SwiftDocsTests.swift
```diff
@@ -2,19 +2,32 @@ import Foundation
...
                        file: StaticString = #file,
                        line: UInt = #line) {
+    let jsonString = String(describing: jsonString)
+
     // Strip out fixtures directory since it's dependent on the test machine's setup
-    let escapedFixturesDirectory = rootDirectory.replacingOccurrences(of: "/", with: "\\/")
-    let jsonString = String(describing: jsonString).replacingOccurrences(of: escapedFixturesDirectory, with: "")
+    let escapedFixturesDirectory = URL(fileURLWithPath: rootDirectory).standardizedFileURL.withUnsafeFileSystemRepresentation {
+        String(cString: $0!) + FileManager.pathSeparator
+    }.replacingOccurrences(of: "\\", with: "\\\\").replacingOccurrences(of: "/", with: "\\/")
```
jpsim, line 23: Same here, could you please use raw string literals to avoid having to escape `\` which adds a lot of noise?

Tests/SourceKittenFrameworkTests/ModuleTests.swift
compnerd: I like the idea of the helper.  However, I do wonder if we should replicate `/usr/bin/env`?  I think that we want that for Windows as well, but does not have that.  We could explicitly use `CreateProcess` but that is a lot of complexity.

Source/SourceKittenFramework/library_wrapper.swift
compnerd: Sure, can move the `WindowsError` into its own file.  Conforming to `Error` seems like a good idea.

Tests/SourceKittenFrameworkTests/ClangTranslationUnitTests.swift
compnerd: I don't think that the `unsafeSystemPath` makes sense, perhaps `FileUtils.nativePathRepresentation`?  The `withUnsafeFileSystemRepresentation` is because the closure receives a `UnsafePointer<CChar>`, but we are transacting in `String` at this layer.

Source/SourceKittenFramework/library_wrapper.swift
compnerd: Done with #784

Tests/SourceKittenFrameworkTests/ModuleTests.swift
jpsim: The equivalent of `/usr/bin/env` on windows would certainly be helpful.

Maybe some abstraction should live in https://github.com/apple/swift-system? In any case I’m happy to consider it out of scope for initial Windows support in SourceKitten.

compnerd: I put together https://github.com/compnerd/Finder and we can simply do `which("git")` and `which("swiftc")` - it respects `PATHEXT` which means that we don't even need to worry about the suffix.

PR #779, opened 2023-06-14
Add partial Windows support

jpsim:
This cherry-picks the following commits from https://github.com/jpsim/SourceKitten/pull/769:

- SourceKittenFramework: use `free` rather than `UMRP.deallocate`
- SourceKittenFramework: treat Windows more similar to Darwin
- SourceKittenFramework: add Windows spelling for sourcekitd
- SourceKittenFramework: adjust import for Windows
- SourceKittenFramework: support Windows SDK for Swift code
- SourceKittenFramework: adjust source path computation
- sourcekitten: import the Windows C library
- Tests: disable syntaxtree tests on 5.9
- Tests: add another version marker for Swift 5.9 (next?)
- Tests: prefer `XCTUnwrap` over forced unwrapping
- Tests: handle platform specific golden results
- Tests: add the known set of strings for Windows

This isn't enough to get SourceKitten building and running on Windows, but rather are the building blocks that should be safe to merge as we work out the remaining issues.


compnerd:
Would you be open to dropping the topmost commit from the initially committed changes?  I'd like to see if we can do something about that which avoids a special case for Windows.  At least a discussion around how to best handle that and if we cannot find anything, we can do this.

compnerd:
The sourcekitStrings changes (https://github.com/jpsim/SourceKitten/pull/779/commits/057d71c34467d384e0a71d0e2736c08930c29511 and https://github.com/jpsim/SourceKitten/pull/779/commits/d1ba00fb521b21653933709dff0563dc7c33b9c6)

This would require that we have a separate hardcoded list for Windows and a dynamic list for non-Windows.  I feel like that might make things confusing.  I wonder if we can figure out how to get that set dynamically or statically everywhere.

jpsim:
> This would require that we have a separate hardcoded list for Windows and a dynamic list for non-Windows.

I'm ok with this for now. If you find a way to get this dynamically for Windows in the future, then we can replace this approach then, which I agree would certainly be nice.

---

I'll merge this now that CI is passing again with the currently supported platforms. Then you can rebase #769 with a more focused set of changes.

PR #781, opened 2023-06-14
Add Swift 5.8 Support

jpsim:
- Add type hints to resolve Swift 5.8 warnings
- Add 5.8 test fixtures
- Adjust compiler versions in tests
- Add `source.lang.swift.syntaxtype.operator` syntax kind
- Add `source.lang.swift.decl.macro` declaration kind
- Add new declaration attribute kinds

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ##### Enhancements
 
-* None.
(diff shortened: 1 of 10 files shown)


jpsim:
I'll add Swift 5.8 CI jobs in a follow-up PR.

PR #784, opened 2023-06-14
SourceKittenFramework: largely port library loading for Windows

compnerd:
This adds Windows specific handling for dynamic symbol resolution into a library.  The `dlopen` and `dlsym` APIs are not portable as they are part of POSIX rather than the C standard.  Use the Windows equivalents of `LoadLibraryW` and `GetProcAddress`.

TODO: this requires further refinement to detect the Swift toolchain installation.

### DIFF
--- a/Source/SourceKittenFramework/WindowsError.swift
+++ b/Source/SourceKittenFramework/WindowsError.swift
@@ -1,0 +1,32 @@
+#if os(Windows)
+import WinSDK
+
+@_transparent
+internal func MAKELANGID(_ p: WORD, _ s: WORD) -> DWORD {
+    return DWORD((s << 10) | p)
+}
+
+struct WindowsError {
+    public let code: DWORD
+}
+
+extension WindowsError: Error {
+    var localizedDescription: String {
+        let dwFlags: DWORD = DWORD(FORMAT_MESSAGE_ALLOCATE_BUFFER)
+                           | DWORD(FORMAT_MESSAGE_FROM_SYSTEM)
+                           | DWORD(FORMAT_MESSAGE_IGNORE_INSERTS)
+        let dwLanguageId: DWORD =
+            MAKELANGID(WORD(LANG_NEUTRAL), WORD(SUBLANG_DEFAULT))
+
+        var buffer: UnsafeMutablePointer<WCHAR>?
+        let dwResult = withUnsafeMutablePointer(to: &buffer) {
+            $0.withMemoryRebound(to: WCHAR.self, capacity: 2) {
+                FormatMessageW(dwFlags, nil, code, dwLanguageId, $0, 0, nil)
+            }
+        }
+        guard dwResult > 0, let message = buffer else { return "Unknown error" }
+        defer { LocalFree(message) }
+        return String(decodingCString: message, as: UTF16.self)
+    }
+}
+#endif
--- a/Source/SourceKittenFramework/WindowsError.swift
+++ b/Source/SourceKittenFramework/WindowsError.swift
@@ -7,14 +7,14 @@ internal func MAKELANGID(_ p: WORD, _ s: WORD) -> DWORD {
 }
 
 struct WindowsError {
-    public let code: DWORD
+    let code: DWORD
 }
 
 extension WindowsError: Error {
     var localizedDescription: String {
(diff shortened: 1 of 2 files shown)


Source/SourceKittenFramework/WindowsError.swift
```diff
@@ -0,0 +1,32 @@
+#if os(Windows)
+import WinSDK
+
+@_transparent
```
jpsim, line 4: Why `@_transparent` instead of `@inlinable`? According to this doc I don't see why this should be transparent: https://github.com/apple/swift/blob/main/docs/TransparentAttr.md

compnerd: 2 reasons:
1) less divergence from the original source of this code (https://github.com/compnerd/swift-win32)
2) This really is a macro in C and the `@_transparent` mimics the behaviour more faithfully:

```c
#define MAKELANGID(p, s) ((s) << 10 | (p))
```

I can change it if you feel strongly.

jpsim: I don’t feel strongly, just trying to understand the motivation. Thanks for sharing.

Source/SourceKittenFramework/WindowsError.swift
```diff
@@ -0,0 +1,32 @@
...
+
+struct WindowsError {
+    public let code: DWORD
+}
+
+extension WindowsError: Error {
+    var localizedDescription: String {
+        let dwFlags: DWORD = DWORD(FORMAT_MESSAGE_ALLOCATE_BUFFER)
+                           | DWORD(FORMAT_MESSAGE_FROM_SYSTEM)
+                           | DWORD(FORMAT_MESSAGE_IGNORE_INSERTS)
```
jpsim, line 17: This should fix one of the SwiftLint violations:

```suggestion
        let dwFlags = DWORD(FORMAT_MESSAGE_ALLOCATE_BUFFER)
                    | DWORD(FORMAT_MESSAGE_FROM_SYSTEM)
                    | DWORD(FORMAT_MESSAGE_IGNORE_INSERTS)
```

Source/SourceKittenFramework/WindowsError.swift
```diff
@@ -0,0 +1,32 @@
+#if os(Windows)
+import WinSDK
+
+@_transparent
+internal func MAKELANGID(_ p: WORD, _ s: WORD) -> DWORD {
+    return DWORD((s << 10) | p)
+}
+
+struct WindowsError {
+    public let code: DWORD
```
jpsim, line 10: This declaration isn't actually public because its parent declaration is internal, leading to a SwiftLint violation.

```suggestion
    let code: DWORD
```

Source/SourceKittenFramework/library_wrapper.swift
```diff
@@ -1,16 +1,33 @@
...
 
 // MARK: - Shared Types & Functions
 
 struct DynamicLinkLibrary {
-    fileprivate let handle: UnsafeMutableRawPointer
+#if os(Windows)
+    typealias HandleType = HMODULE?
+#else
+    typealias HandleType = UnsafeMutableRawPointer
+#endif
```
jpsim, line 13: Nit: Generally typealiases in Swift aren't suffixed with `Type`

```suggestion
#if os(Windows)
    typealias Handle = HMODULE?
#else
    typealias Handle = UnsafeMutableRawPointer
#endif
```

Source/SourceKittenFramework/library_wrapper.swift
```diff
@@ -1,16 +1,33 @@
...
 
 struct DynamicLinkLibrary {
-    fileprivate let handle: UnsafeMutableRawPointer
+#if os(Windows)
+    typealias Handle = HMODULE?
+#else
+    typealias Handle = UnsafeMutableRawPointer
+#endif
+
+    fileprivate let handle: HandleType
```
compnerd, line 15:
```suggestion
    fileprivate let handle: Handle
```

jpsim:
> TODO: this requires further refinement to detect the Swift toolchain installation.

I’m fine if you do this as a follow-up.

compnerd:
Yeah, I think that we should address that as a follow up.  The current plan to change the install location means that this path is going to need to be dynamic.

jpsim (approved):

PR #785, opened 2023-06-15
Tests: normalise the fixtures path for Windows

compnerd:
Ensure that we normalise the path when we substitute the path into the results to properly remove the basename of the fixture.

### DIFF
--- a/Tests/SourceKittenFrameworkTests/SwiftDocsTests.swift
+++ b/Tests/SourceKittenFrameworkTests/SwiftDocsTests.swift
@@ -7,14 +7,24 @@ func compareJSONString(withFixtureNamed name: String,
                        rootDirectory: String = fixturesDirectory,
                        file: StaticString = #file,
                        line: UInt = #line) {
+    let jsonString = String(describing: jsonString)
+
     // Strip out fixtures directory since it's dependent on the test machine's setup
+#if os(Windows)
+    let escapedFixturesDirectory = URL(fileURLWithPath: rootDirectory)
+        .standardizedFileURL
+        .withUnsafeFileSystemRepresentation { String(cString: $0!) }
+        .replacingOccurrences(of: #"\"#, with: #"\\"#)
+        .replacingOccurrences(of: "/", with: #"\/"#)
+#else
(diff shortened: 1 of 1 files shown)


jpsim:
Approach looks ok, I'll merge when you fix the remaining issues causing tests to fail.

compnerd:
@jpsim any idea why this would be failing?

jpsim:
> any idea why this would be failing?

It's adding a trailing `/` on macOS/Linux which wasn't there before.

This fixes the macOS/Linux tests:

```swift
// Strip out fixtures directory since it's dependent on the test machine's setup
let escapedFixturesDirectory = URL(fileURLWithPath: rootDirectory)
    .standardizedFileURL
    .withUnsafeFileSystemRepresentation { String(cString: $0!) }
    .replacingOccurrences(of: #"\"#, with: #"\\"#)
    .replacingOccurrences(of: "/", with: #"\/"#)
let escapedJSONString = jsonString.replacingOccurrences(of: escapedFixturesDirectory, with: "")
```

Is it ok to have the Windows paths skip adding the trailing path separator?

compnerd:
```
-  "Subscript.swift" : {
+  "\/Subscript.swift" : {
```

This diff makes me think that the trailing suffix is important?

jpsim:
We can just do a separate codepath for Windows for the time being:

```swift
    // Strip out fixtures directory since it's dependent on the test machine's setup
#if os(Windows)
    let escapedFixturesDirectory = URL(fileURLWithPath: rootDirectory)
        .standardizedFileURL
        .withUnsafeFileSystemRepresentation { String(cString: $0!) }
        .replacingOccurrences(of: #"\"#, with: #"\\"#)
        .replacingOccurrences(of: "/", with: #"\/"#)
#else
    let escapedFixturesDirectory = rootDirectory.replacingOccurrences(of: "/", with: #"\/"#)
#endif
    let escapedJSONString = jsonString.replacingOccurrences(of: escapedFixturesDirectory, with: "")
```

Adjust that for whatever you need on the Windows side.

jpsim (approved):

PR #786, opened 2023-06-24
Tests: escape the path separator for Windows

compnerd:
`\` must be escaped for valid JSON.  Ensure that we properly escape the character and form the reference string with the proper representation of the path.


compnerd:
@jpsim not sure what to do here - the problem I suspect involves the behaviour of symbolic links

PR #801, opened 2024-01-07
Added a check for MacOS Catalyst since it should go the IOS path

chotchki:
MacOS Catalyst is an iOS target that runs on MacOS. Enhancing this call so the SwiftLint Plugin will work on a catalyst build target


jpsim (approved):
Thanks!

PR #803, opened 2024-02-03
fix(linux): do not assume the location of `which`

judemille:
In this PR, it is invoked, like other commands, using `/usr/bin/env`.


jpsim (approved):
Looks good, thank you

PR #808, opened 2024-02-15
Bump cocoapods to 1.15.2

marcosgriselli:
The GitHub Actions macOS 13 runner image got updated 2 days ago and now [Xcode 15.0.1 is the default selected Xcode](https://github.com/actions/runner-images/pull/9292/files#diff-49d5008075920b3466eff9e2bb3f4a933cca7733524c0153fd761cf9b6fc4fd4R169). This change started to fail the [`pod lib lint` action](https://github.com/jpsim/SourceKitten/actions/runs/7837560900/job/21607163170) due to [this cocoapods bug in Xcode 15+](https://github.com/CocoaPods/CocoaPods/pull/12009).

After updating cocoapods the command passed on the following Xcode versions: 
✅ Xcode 14.3.1
✅ Xcode 15.2.0
✅ Xcode 15.3.0 Beta 3

PR #809, opened 2024-02-15
Update Xcode versions used on actions

marcosgriselli:
- [x] Update the Xcode versions used in GitHub Actions.
- [x] Add an explicit version to the `pod lib lint` action to prevent image runner updates from breaking the job (https://github.com/jpsim/SourceKitten/pull/808)


johnfairh:
Good idea but the tests will not pass on Swift 5.9, continuing this in #810

PR #810, opened 2024-02-26
Support for Swift 5.9

johnfairh:
First step to adding Swift 5.10 support....

* Update CI to use Swift 5.9 as per #809
* Include gem update as per #808 to get cocoapods workaround for Xcode 15
* Fixtures update for Swift 5.9 SourceKit
    * New attributes
    * New cursorinfo `key.reusingastcontext` field
    * Workaround for Windows code vs. SourceKit 5.9


Source/SourceKittenFramework/SwiftDeclarationAttributeKind.swift
```diff
@@ -71,7 +71,7 @@ public enum SwiftDeclarationAttributeKind: String, CaseIterable {
     case setterPublic = "source.decl.attribute.setter_access.public"
     case setterOpen = "source.decl.attribute.setter_access.open"
     case optimize = "source.decl.attribute._optimize"
-    case consuming = "source.decl.attribute.__consuming"
+    case __consuming = "source.decl.attribute.__consuming"
```
johnfairh, line 74: In conjunction with the new 'real' `consuming` attribute this rename is a bit dodgy, open to suggestions...

jpsim (approved):
Thanks! I know Xcode / Swift updates are tedious but I appreciate the work here.

PR #811, opened 2024-02-29
Support Swift 5.10

johnfairh:
(e: hit the button too soon...)

Main thing is that SwiftPM's yaml file changed format which needs some actual coding.  Fixture updates minimal:

cursorinfo
* Replacement of `Foo.Protocol` with `Foo.Type` for protocol metatypes
* @_eagerMove appearing on `Swift.String` (no, me neither)

objc
* Ubiquitous `any` for Swift versions of APIs eg. `any NSCopying`

Code completion
* `any` in various fields
* a load of completions have disappeared but they all seem to be for `Double` leaving just the `Int` ones which seems like an improvement


Source/SourceKittenFramework/Module.swift
```diff
@@ -202,3 +192,67 @@ private extension Collection where Element == XcodeBuildSetting {
         return lazy.compactMap(getterClosure).first
     }
 }
+
+// MARK: Yams.Node helpers for SwiftPM
+
+// The yaml structure changed in Xcode 15.3 / SwiftPM 5.10.  This extension decodes both formats.
+private extension Node {
```
SimplyDanny, line 199: You can define Swift version `fiveDotTen` in [SwiftVersion](https://github.com/jpsim/SourceKitten/blob/main/Source/SourceKittenFramework/SwiftVersion.swift) and pick the correct selection path depending on this version here instead of trying both.

johnfairh: I don't think so: `SwiftVersion` is about the version of Swift providing SourceKit whereas this code is about some version of Swift whose SwiftPM created a `debug.yaml`.

johnfairh:
Unexpected struggle getting this to work:  Xcode 15.3 has an undocumented feature in `xcodebuild` and other places where environment variables from the invoking environment are discarded.  This includes `DEVELOPER_DIR`.

So when sourcekitten tries to fork `xcodebuild`, it gets the system default version (15.0.1 in the GHA image) which does not work.

Changed CI to use the `xcode-select` method instead.

[Dev forums](https://forums.developer.apple.com/forums/thread/749185), FB linked from there.

(Presumably intended as a hermetic testing improvement but shame on them for not mentioning it, or coordinating with SwiftPM, or…)

johnfairh:
Longer than I meant to take getting back to this.

Apple has [confirmed](https://forums.developer.apple.com/forums/thread/749185) that this env. var thing is working as designed -- sounds like the alternative to `xcode-select` is to set both `DEVELOPER_DIR` _and_ `TEST_RUNNER_DEVELOPER_DIR`.  I think using `xcode-select` is probably fine & the right advice for consumers.

@jpsim fyi, will merge this in a few days.

jpsim:
Thanks for driving all this, @johnfairh!

PR #816, opened 2024-06-29
Adds support for rules_swift 2.x.

luispadron:
This does force users using this version of sourcekitten to also use rules_swift because of the `swift_c_module` removal.

Outstanding issues:

- [x] `max_compatibility_level` needs to be updated for deps that use rules_swift 
- [x] Need a real `2.0.0` release before we can move of the release candidate

Requires the following:

- https://github.com/jpsim/Yams/pull/420

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -10,7 +10,7 @@ jobs:
   MacOS:
     strategy:
       matrix:
-        xcode_version: ['14.3.1', '15.2', '15.3']
+        xcode_version: ['15.2', '15.3']
     runs-on: macos-14
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


jpsim:
Yams 5.1.3 has now been released: https://github.com/jpsim/Yams/releases/tag/5.1.3

BCR PR: https://github.com/bazelbuild/bazel-central-registry/pull/2404

luispadron:
Thank you! The remaining module would be: https://github.com/drmohundro/SWXMLHash which i think was manually added to the registry as i dont see any bazel files in that repo

.github/workflows/bazel.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   MacOS:
     strategy:
       matrix:
-        xcode_version: ['14.3.1', '15.2', '15.3']
```
luispadron, line 13: The 14.3.1 job was failing on Bazel because of diff in fixtures. I'm not able to install that version on my OS. If we're okay with bumping this this should be ready now

jpsim: Yeah that's a fairly old version now, fine to remove from CI

jpsim (approved):

PR #830, opened 2024-12-14
Use SWXMLHash 8.1.0

SimplyDanny:
Requires https://github.com/bazelbuild/bazel-central-registry/pull/3416.


johnfairh:
So SWXMLHash [isn't supporting CocoaPods from its 8.x.y release](https://github.com/drmohundro/SWXMLHash/releases/tag/8.0.0).  If SourceKitten really needs v8 of this to build/work then I think we have to drop CocoaPods as well?

SimplyDanny:
> So SWXMLHash [isn't supporting CocoaPods from its 8.x.y release](https://github.com/drmohundro/SWXMLHash/releases/tag/8.0.0). If SourceKitten really needs v8 of this to build/work then I think we have to drop CocoaPods as well?

I cannot remember anymore why I updated this. It doesn't seem to have anything that is required on my end. We can just close this until we see a need.

johnfairh:
Great, best possible answer!

PR #832, opened 2025-03-21
SPM: Explicit declarations for external dependencies

ailinykh:
When I tried to build my project using Xcode 16, I got an error:
```bash
error: 'sourcekitten': dependency 'Yams' in target 'SourceKittenFramework' requires explicit declaration; reference the package in the target dependency with '.product(name: "Yams", package: "yams")'
make: *** [build] Error 1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,18 @@
+## Main
+
+#### Breaking
+
+* None.
+
+#### Enhancements
+
+* Fully specify package dependencies for wider compatibility.  
+  [Anthony Ilinykh](https://github.com/ailinykh)
+
+#### Bug Fixes
+
+* None.
+
 ## 0.37.0
 
 #### Breaking
--- a/Package.swift
+++ b/Package.swift
@@ -32,8 +32,8 @@ let package = Package(
             dependencies: [
                 "Clang_C",
                 "SourceKit",
-                "SWXMLHash",
-                "Yams",
+                .product(name: "SWXMLHash", package: "SWXMLHash"),
+                .product(name: "Yams", package: "Yams"),
             ]
         ),
         .testTarget(


ailinykh:
Here what is my setup looks like:
```mermaid
graph TD;
    MyProject-->MyPrivateProject;
    MyPrivateProject-->Yams;
    SourceKitten-->Yams;
    MyProject-->SourceKitten;
```
so I suppose that SPM can't make a decision what `Yams` to use when it comes to build `SourceKittenFramework`. Should it checkout a new one, or it could rely to `MyPrivateProject` dependency

SourceKitten is 0.73.0
Yams is  5.3.1 (in both cases)

ailinykh:
I made more precise research and what I've found:
1. Error message comes from [ModulesGraph.swift](https://github.com/swiftlang/swift-package-manager/blob/6c179de92d5b14040f0171226c4a97c3a8a74229/Sources/PackageGraph/ModulesGraph.swift#L320)
2. The `productDependencyMissingPackage` error thrown from [ModulesGraph+Loading.swift](https://github.com/swiftlang/swift-package-manager/blob/6c179de92d5b14040f0171226c4a97c3a8a74229/Sources/PackageGraph/ModulesGraph%2BLoading.swift#L706)
3. The [comment above](https://github.com/swiftlang/swift-package-manager/blob/6c179de92d5b14040f0171226c4a97c3a8a74229/Sources/PackageGraph/ModulesGraph+Loading.swift#L688) says:

> // Starting in 5.2, and module-based dependency, we require module product dependencies to
   // explicitly reference the package containing the product, or for the product, package and
   // dependency to share the same name. We don't check this in manifest loading for root-packages so
   // we can provide a more detailed diagnostic here.

Therefore I suspect that the root cause is target name `Yams` matches the package name `yams` and SPM can't resolve it

johnfairh:
Well, neither of those ideas speak to why your project is the only one affected, but the change is good & will hopefully help others in the future too -- thank you for the PR!

PR #834, opened 2025-06-11
Support rules_swift 3.0

luispadron:
Update `max_compatibility_level` for rules_swift and update dependencies to do the same.


jpsim:
Please revert the SwiftPM package manifest changes for maximum compatibility with consumers. Otherwise looks good!

luispadron:
Done!

PR #835, opened 2025-06-15
Update dependencies and consume them from BCR

SimplyDanny:
All dependencies are now available on BCR and thus don't require manual builds.


jpsim:
Thanks! Do you need me to tag a release with this change?

SimplyDanny:
> Thanks! Do you need me to tag a release with this change?

That would be awesome!

PR #845, opened 2026-06-29
Swift 6.4 support

johnfairh:
Update for Swift 6.4

Few basic changes using Xcode 27b2 / swiftlang-6.4.0.23.5

Then a bigger change to deal with SwiftPM 6.4, which switches the build backend to "Swift Build".
Sadly this no longer produces the debug.yaml file that we used to get compiler arguments for SourceKit.

This PR adds an Xcode-like method for SwiftPM -- build the package with a "-v" flag, spot the command to build the module, and parse out the compiler arguments and source files.

Very little new code for this, reuses the Xcode arg-parsing routines, but moving all the SwiftPM-specific stuff into a new file, refactors in separate commits.

WIP - see how the beta train goes over the summer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 * New syntax, attribute and declaration kinds introduced in Swift 6.4.  
   [John Fairhurst](https://github.com/johnfairh)
 
+* Docs generation and code completion with Swift PM 6.4.  
+  [John Fairhurst](https://github.com/johnfairh)
+
 #### Bug Fixes
 
 * None.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 #### Enhancements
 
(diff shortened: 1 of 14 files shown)


johnfairh:
So the Swift PM 6.4 design point has remained, the debug.yaml is gone.  They did fix the bug I reported that stopped the 'parse the CLI' thing from working, so this is a workable place to publish.

I did work out a 'no recompile' [method](https://github.com/johnfairh/SourceKitten/blob/f471453d051370a0f98583f0bcb61e9aeff8f294/Source/SourceKittenFramework/SwiftPMTaskReader.swift) that reads the custom-msgpack task store, but it's more code than I'm comfortable maintaining in sourcekitten and very fragile sharing magic numbers and binary layouts with llbuild or some such - not to mention I grabbed some basic msgpack code of uncertain license.

